PR #3, opened 2025-02-06
opens possibility to build images of various Sonic versions

kjezek:
This PR update `Dockerfile` and build process to allow for building multiple versions of the Sonic client. 

It parametrises the Dockerfile to checkout a specified version of Sonic client from git. It means that the client version from the `client/` directory is no more used. 

`Makefile` was updated to build Docker images from all Sonic tags we have at the moment. 

This is a first step to run various versions of the client in an experiment.  Next steps will be:
1. unlink the `client/` dir
2. extend scenario configuration to define an image
3. possibly make startup scripts also configurable to accomodate changes in start-up procedure of the client

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,7 +21,7 @@
 #
 FROM golang:1.22 AS client-build
 
-# Download Sonic dependencies from the default bracnch first to cache them.
+# Download Sonic dependencies from the default branch first to cache them.
 # We assume tags/branches do not change majority of dependencies.
 RUN git clone https://github.com/0xsoniclabs/sonic.git /client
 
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -11,27 +11,44 @@
...
+# Its caches the dependencies first, so that the build is faster.
+#
+# It checks out the required version of the client, and builds it.
+#
+FROM golang:1.22 AS client-build
 
-WORKDIR /client
-COPY client/go.mod ./go.mod
-RUN go mod download
+# Download Sonic dependencies from the default bracnch first to cache them.
```
HerbertJordan, line 24:
```suggestion
# Download Sonic dependencies from the default branch first to cache them.
```

HerbertJordan (approved):
Looks good, thanks!

PR #4, opened 2025-02-06
Kjezek/configurable image name

kjezek:
This PR propagates image name from scenario configuration. The image name is used when starting the container, which allows for running various client versions. 

An example configuration is presented.

### DIFF
--- a/driver/executor/executor.go
+++ b/driver/executor/executor.go
@@ -208,6 +208,11 @@ func scheduleNodeEvents(node *parser.Node, queue *eventQueue, net driver.Network
 	}
 	nodeIsCheater := false
 
+	image := driver.ClientDockerImageName
+	if node.Client.ImageName != "" {
+		image = node.Client.ImageName
+	}
+
 	for i := 0; i < instances; i++ {
 		name := fmt.Sprintf("%s-%d", node.Name, i)
 		var instance = new(driver.Node)
@@ -218,6 +223,7 @@ func scheduleNodeEvents(node *parser.Node, queue *eventQueue, net driver.Network
 			func() error {
 				newNode, err := net.CreateNode(&driver.NodeConfig{
 					Name:      name,
+					Image:     image,
 					Validator: nodeIsValidator,
 					Cheater:   nodeIsCheater,
 				})
--- a/driver/executor/executor.go
+++ b/driver/executor/executor.go
@@ -208,7 +208,7 @@ func scheduleNodeEvents(node *parser.Node, queue *eventQueue, net driver.Network
 	}
 	nodeIsCheater := false
 
-	image := driver.ClientDockerImageName
+	image := driver.DefaultClientDockerImageName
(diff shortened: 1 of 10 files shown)


scenarios/test/various_client_versions.yml
```diff
@@ -0,0 +1,55 @@
...
+      imagename: "sonic:v2.0.1"
+
+  - name: validator-A
+    start: 10
+    instances: 1
+    client:
+      imagename: "sonic:v2.0.0"
+      type: validator
+
+# In the network, there is a single application producing a constant load.
```
HerbertJordan, line 35: nit: there are actually 3 applications ..

driver/network.go
```diff
@@ -26,6 +26,8 @@ import (
 
 //go:generate mockgen -source network.go -destination network_mock.go -package driver
 
+const ClientDockerImageName = "sonic"
```
HerbertJordan, line 29: Which version is taken if the name is just `sonic` like in this case?

HerbertJordan, line 29: nit: please document this constant; also, maybe rename to `DefaultClientDockerImage` ?

driver/executor/executor.go
```diff
@@ -208,6 +208,11 @@ func scheduleNodeEvents(node *parser.Node, queue *eventQueue, net driver.Network
 	}
 	nodeIsCheater := false
 
+	image := driver.ClientDockerImageName
+	if node.Client.ImageName != "" {
+		image = node.Client.ImageName
+	}
```
HerbertJordan, line 214: nit: since you are enforcing a default value here anyway, many changes setting the image name for tests might not be necessary;

HerbertJordan (approved):
Looks good, thanks for adding this capability!

driver/network.go
kjezek: in docker you may build an image with or without the version tag. The tagged image is supposed to be unchangeable (albeit not enforced for a local build), while the untagged version is an ongoing version, like `main/master` in git.   Sometimes `<image>:latest` is used as well

driver/executor/executor.go
kjezek: This default is set for `driver.NodeConfig` only, but in the tests I configure `opera.OperaNodeConfig` - which is a layer below the executor. 

This layer does not have a default value, because it is not directly created from the configuration, where defaults are set.

driver/network.go
kjezek: ok renamed

HerbertJordan (approved):

driver/network.go
HerbertJordan: So, just to check that I got it, `sonic` is the image to chose if one wants to run the `main` branch, right?

kjezek: yes, it is correct - owing the fact you run `make` (or `make build-sonic-docker-image`) to have an up to date checkout and build of Sonic `main`.

PR #5, opened 2025-02-10
extracts sonic tools into its own submodule

kjezek:
This PR extracts `normatools` package to its own submodule. 

The original design had an issue that `norma` had a dependency to `Sonic` client. This dependency was used for generating startup parameters for running the client. 

Norma needs to be shifted to allow for running various versions of the client in one network. For this reason Norma itself cannot be bound to a particular version of the client. 

Instead, the idea is to create this submodule, which will only refer to a certain client version. If another incompatible version of the client is needed to be supported, a new submodule adapted to this new version of the client may be created.  

Unfortunately, one dependency to the Sonic client still remains:
* https://github.com/0xsoniclabs/norma/blob/main/load/app/validator.go
* This code executes SFC and it is unclear how to remove this dependency


rpl-ffl (approved):
As per our discussion, the changes here looks good to me.
All SFC interaction are to be refactored into `genesis` and `genesistool` as appropriate - in another PR.

PR #7, opened 2025-02-13
configures variable startup validators

kjezek:
This PR extends scenario configuration of network bootstrap validators. 

Before this change, the network could be configured for the number of validators that were started to bootstrap the network. This PR extends this configuration that validators can be configured running different versions of the client. 

The configuration defines a new validators map where each client version can be defined with the number of instances it runs.


driver/network.go
```diff
@@ -119,3 +122,53 @@ type ApplicationConfig struct {
...
+		Instances: instances,
+		ImageName: imageName,
+	}
+}
+
+type Validators struct {
+	Validators []Validator
+}
+
+// NewDefaultValidators creates a new Validators with a single validator using
```
HerbertJordan, line 154: nit: this comment just ends without completing the sentence.

driver/parser/parser.go
```diff
@@ -69,20 +69,20 @@ func (s *Scenario) GetMaxEpochGas() uint64 {
...
 		return *s.RoundTripTime
 	}
 	return 0
 }
 
+// Validator is a configuration for a group of network start-up validators.
+type Validator struct {
+	Name      string
+	Instances *int       `yaml:",omitempty"` // nil is interpreted as 1
+	Client    ClientType `yaml:",omitempty"`
```
HerbertJordan, line 83: nit: an optional image name may be sufficient here, unless you plan to extend the `ClientType` with additional fields in the future.

driver/network.go
```diff
@@ -29,6 +29,9 @@ import (
 // DefaultClientDockerImageName is the name of the docker image to use for clients.
 const DefaultClientDockerImageName = "sonic"
 
+// DefaultValidator is a default configuration for a single validator.
+var DefaultValidator = NewDefaultValidators(1)
```
HerbertJordan, line 33: nit:
```suggestion
var DefaultValidators = NewDefaultValidators(1)
```

driver/network/local/local.go
```diff
@@ -111,23 +111,28 @@ func NewLocalNetwork(config *driver.NetworkConfig) (*LocalNetwork, error) {
...
-			nodeConfig := node.OperaNodeConfig{
-				ValidatorId:   &validatorId,
-				Image:         driver.DefaultClientDockerImageName,
-				NetworkConfig: config,
-				Label:         fmt.Sprintf("_validator-%d", validatorId),
-			}
-			net.validators[i], errs[i] = net.createNode(&nodeConfig)
-		}()
+	var idx int
+	for _, validator := range config.Validators.Validators {
```
HerbertJordan, line 118: nit: you might consider using `[]Validator` instead of the `Validators` struct wrapping it; I think you could even define `Validators` like this:
```
type Validators []Validator

func (vs Validators) GetNumValidators() int {
   // ...
}
```
to kinda have the best of both worlds.

scenarios/load_tests/slope.yml
```diff
@@ -2,8 +2,12 @@
 name: Slope Load Test
 duration: 120
 
-# The number of static validators present from begin to end.
-num_validators: 4
+# Initial validator nodes in the network.
+validators:
+  - name: validator-1
```
HerbertJordan, line 7: nit: this results in names like "validator-1_validator-0", "validator-1_validator-1", ... , "validator-1_validator-3";

Maybe the startup code should just add the counter and this field here should be reduced to "validator", ideally with this being the default value if nothing is set, to obtain the naming "validator-0", "validator-1", ...

driver/network_test.go
```diff
@@ -0,0 +1,140 @@
+package driver
+
+import (
+	"github.com/0xsoniclabs/norma/driver/parser"
+	"golang.org/x/exp/slices"
+	"testing"
+)
+
+var onet int = 1
```
HerbertJordan, line 9:
```suggestion
var one int = 1
```

HerbertJordan (dismissed):
Looks good, thanks for adding this feature!
All comments are of a cosmetic nature, feel free to ignore them.

scenarios/eval/scalability.yml
```diff
@@ -3,7 +3,13 @@
 
 name: Scalability
 duration: 600
-num_validators: 1   # to be overriden by command line flag
+
+# Initial validator nodes in the network.
+validators:
+  - name: validator-1
+    instances: 1
```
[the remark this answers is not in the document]
kjezek: ok, will remove except an example scenario

driver/parser/check.go
```diff
@@ -289,8 +289,17 @@ func checkTimeInterval(start, end *float32, duration float32) error {
 
 // checkValidatorConstraints makes sure that there is correct number of validators.
 func (s *Scenario) checkValidatorConstraints() error {
-	if s.NumValidators != nil && *s.NumValidators < 1 {
-		return fmt.Errorf("invalid number of validators: %d < 1", *s.NumValidators)
+	var numValidators int
```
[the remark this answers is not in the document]
kjezek: ok, I will simplify to remove `ClientType` it is useless

driver/network.go
kjezek: right, fixed

driver/parser/parser.go
kjezek: ok, good idea

driver/network/local/local.go
kjezek: ah, good point, how could I forget this definition is possible

scenarios/load_tests/slope.yml
kjezek: right, removed the `_validator` part

driver/network.go
```diff
@@ -29,6 +29,9 @@ import (
 // DefaultClientDockerImageName is the name of the docker image to use for clients.
 const DefaultClientDockerImageName = "sonic"
 
+// DefaultValidator is a default configuration for a single validator.
```
HerbertJordan, line 32:
```suggestion
// DefaultValidators is a default configuration for a single validator.
```

release_testing/c1.RpcRequests.yml
```diff
@@ -8,8 +8,10 @@
 name: C1
 duration: 300 # 5 minutes
 
-# backward compatibility, to be removed.
-num_validators: 4
+# Initial validator nodes in the network.
+validators:
+  - name: validator-1
+    instances: 4
```
HerbertJordan, line 14: If I understand correctly, this will result in `validator-1-0` to `validator-1-3`.
How about making the name `validator` the default, such that a scenario file like
```
validators:
  - instances: 4
```
is all you need to define here and in most other cases?
This would result in the names `validator-0` to `validator-3`

scenarios/long.yml
```diff
@@ -7,8 +7,9 @@ name: Long-Running Network
 # The duration of the scenario's runtime, in seconds.
 duration: 600
 
-# The number of validator nodes in the network.
-num_validators: 1
+# Initial validator nodes in the network.
+validators:
+  - name: validator-1
```
HerbertJordan, line 12: a thought: If the `valdiator` name gets a default, then this would be
```suggestion
validators:
  - instances: 1
```
Or it might even be omitted if you define an empty list in the config to be defaulted to a single validator.

scenarios/test/various_validator_versions.yml
```diff
@@ -0,0 +1,44 @@
...
+validators:
+  - name: validator-1
+  - name: validator-2
+    imagename: "sonic:v2.0.2"
+  - name: validator-3
+    instances: 2
+    imagename: "sonic:v2.0.1"
+  - name: validator-4
+    instances: 3
+    imagename: "sonic"
```
HerbertJordan, line 21: Since it is the demo for how to use this option, I would suggest to rename the validators to produce nice instance names:
```suggestion
  - name: validator-dev
  - name: validator-v2.0.2
    imagename: "sonic:v2.0.2"
  - name: validator-v2.0.1
    instances: 2
    imagename: "sonic:v2.0.1"
  - name: validator-latest
    instances: 3
    imagename: "sonic"
```

HerbertJordan (approved):
Thanks for factoring in the comments, looks good!

release_testing/c1.RpcRequests.yml
kjezek: ok updated that both `name`, `instances`, and the whole `validators` block can be omitted. Default name `validator` and the number of instances = `1` is used when omitted.  When the whole block is omitted, one default validator is created.

HerbertJordan (dismissed):
Thanks for following up on all those ideas, looking forward to using it.

rpl-ffl (approved):
LGTM, thank you

PR #8, opened 2025-02-14
allows for free-form network rules configuration

kjezek:
This PR allows for defining a list of network rules in scenario configuration. 

This is implemented as a free form key-value list to be easily extensible and it does not have to be touched to support additional rules in the future. 

The way of old configuration was removed, it was not used anyway.


rpl-ffl (approved):
LGTM. Thank you for the refactoring!

PR #9, opened 2025-02-18
Kjezek/generalise client support

kjezek:
This PR generalises sonic client support. It turns out that features contained in this module are more related to the chain, not a particular client implementation. 

For this reason, this module was generalised and renamed to `genesis`


rpl-ffl (approved):
LGTM, Thank you!

rpl-ffl (approved):
Thank you

PR #10, opened 2025-02-18
create genesis generator

kjezek:
This PR creates a genesis file generator. 

This generator reads configuration from environment variables and creates a json file that can be further consumed by the Sonic client to bootstrap the network.

Sonic json parser is used for generating the file


rpl-ffl (approved):
LGTM, thank you

PR #22, opened 2025-03-04
Kjezek/enable local client build

kjezek:
This PR extends the set of docker images build for the clients of the local one. 

This local instance is added to the project as a git submodule and built as an additional image tagged as `local` .

This image is useful for local development, when the client can be modified, and immediately tested in the Norma network.


rpl-ffl (dismissed):
Thank you!

HerbertJordan (dismissed):
Thanks for adding support for this very convenient development feature!

rpl-ffl (approved):

PR #23, opened 2025-03-04
extends tests to test correct client version has started

kjezek:
This PR extends testing to assess version of Sonic checked out from git. 

The version is stored in a file in the image and printed out when Sonic starts.


rpl-ffl (approved):
Thank you!

PR #25, opened 2025-03-05
updates network rules configuration to split to genesis and updates

kjezek:
This PR extends configuration of network rules. 

It splits definition to two sections: `genesis` and `updates`.

Genesis is applied to the genesis file when starting the nodes while updates contain changes in the rules during the network run.


HerbertJordan (approved):
Nice format for this new feature, thanks!

rpl-ffl (approved):

PR #26, opened 2025-03-05
extends scenario scheduler of network rules updates

kjezek:
This PR extends the scheduler to trigger updates to network Rules.


rpl-ffl (dismissed):
Quick questions, to confirm my understanding. 
This PR focuses on getting the right timing for `ApplyNetworkRules` but this doesn't do anything yet, is that right?

kjezek:
> Quick questions, to confirm my understanding. This PR focuses on getting the right timing for `ApplyNetworkRules` but this doesn't do anything yet, is that right?

Right, it schedules new events to trigger network updates. The updates are not propagated to the network yet. It will follow up.

rpl-ffl (approved):

PR #27, opened 2025-03-06
adds support to generate json with network rules updates

kjezek:
This PR adds support to generate JSON string describing network updates.

It contains a method that gets a pivotal network rules object and the key-value pairs on it inputs. It applies network rules from the key-value pairs to the network rules object. Finally it generates JSON string that contains only items that have changed between the original and new network rules objet.


rpl-ffl (approved):
Thank you!

PR #32, opened 2025-03-12
Switch to use dynamic gas price transactions in load generators

HerbertJordan:
This PR replaces the utilization of `LegacyTx` transactions with `DynamicFee` transactions in the load generators. This change eliminates the need for fetching an estimated gas price from the network before sending a transaction. The `DynamicFee` type of transaction will automatically adjust the gas price to the base-fee level at the time of its execution.

This change was motivated by problems encountered during stress tests, in which load generators started to fail producing transactions because the RPC service used to obtain gas-price predictions could not respond fast enough. By eliminating the need for fetching the gas price, this problem could be resolved.

A positive side effect of this is that load-tests are now more accurately representing the load case of validator nodes as the overhead of serving gas-price estimations is eliminated. Before this change, this included the need to create a client connection and to run a RPC request. This is not what a validator node would typically be expected to be confronted with, as validator nodes are recommended not to expose an RPC interface.

### DIFF
--- a/load/app/app.go
+++ b/load/app/app.go
@@ -17,8 +17,6 @@
 package app
 
 import (
-	"math/big"
-
 	"github.com/0xsoniclabs/norma/driver/rpc"
 	"github.com/ethereum/go-ethereum/core/types"
 )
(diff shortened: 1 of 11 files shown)


kjezek (approved):
thanks, looks good

PR #33, opened 2025-03-12
remove feature configuration from scenario node

kjezek:
This PR removes the `feature` construct from node configuration. It did nothing and was confusing that it actually does not configure anything. In particular, validator nodes are configured differently.


rpl-ffl (approved):
Thank you!

PR #34, opened 2025-03-12
fixes adding dynamic validators during network runtime

kjezek:
This PR fixes adding of dynamic validators. 

It pre-allocates accounts for N validators that can be added during the network runtime.


rpl-ffl (approved):
Thank you!

PR #35, opened 2025-03-13
adds check for unsuported network rules configuratio

kjezek:
This PR adds validation of configured network rules updates and prevents unrecognised configurations. This should help preventing situation like typo in the key value pairs, etc.


rpl-ffl (approved):
Thank you!

PR #39, opened 2025-03-17
creates scenario for hardfork

kjezek:
This PR adds a new scenario that simulates a hardfork.
It updates network rules to enable Allegro in the network of running nodes and verifies that the nodes keep running. 

This PR also extends supported network rule changes. 

There is also an idea to show that an older node fails after network update, but it is not available in Norma at the moment.


rpl-ffl (approved):
Thank you!

PR #43, opened 2025-03-19
Scenario nodes_start_stop now have starting number of validator = 3

rpl-ffl:
This PR sets the number of starting validator of scenario `nodes_start_stop` to 3.
If the number of validator is small, adding multiple validators in a short amount of timespan results in timeout.

Additional tests here:
- Running `nodes_start_stop` sequentially multiple times to ensure that scenario run does not have sporadic issue.
- Nightly baseline scenarios pipeline passes.


kjezek:
I would like to better understand the issue. 

I could understand that the validators loose quorum, causing that txs are not processed. But why the network does not interrupt in time? It could be an issue in Norma

kjezek:
I got this scenario working:
```

# This scenario simulates a network with the various number of nodes.
# It starts and stopes nodes at different times.
# The nodes include validators and non-validator nodes.
# It deploys one application and sends a decent number of transactions.

# The name of the scenario
name: Various Number of Nodes

# The duration of the scenario's runtime, in seconds.
duration: 240

# Shortens epoch duration to reflect changes in validator registrations.
network_rules:
  genesis:
    MAX_EPOCH_DURATION: 5s

# The network scenario to exercise. 
nodes:
  # Change the number of nodes during the scenario.
  - name: validator-A
    start: 20
    instances: 1
    client:
      type: validator

  - name: validator-B
    start: 40
    instances: 1
    client:
      type: validator

  - name: observer-A
    start: 80
    end: 180
    instances: 2

  - name: observer-B
    start: 100
    end: 160
    instances: 1

  - name: observer-C
    start: 190
    end: 230
    instances: 2

  - name: validator-C
    start: 140
    end: 200
    instances: 1
    client:
      type: validator


# In the network, there is a single application producing a constant load.
applications:
  - name: load
    type: counter
    users: 1000           # number of users using the app
    rate:
      constant: 100     # Tx/s

```

rpl-ffl:
@kjezek This PR is now ready for your approval.

- Scenario changed according to suggested comment above.
- Tested for sporadic issues and found none.

kjezek (approved):
thx

PR #45, opened 2025-03-24
improves design of RPC and fix too long timeout

kjezek:
This PR most importantly fixes timeout when waiting for a transaction receipt. It was by accident set to `120 000` seconds. It was reduced to 1 min max.  

Since the related code was not well designed to be tested, the code was reworked, enriched with mockable interfaces, and the updated code tested.


rpl-ffl (approved):
Thank you!

PR #48, opened 2025-03-28
creates scenario to test network rules constraints

kjezek:
This PR adds scenarios to test combinations of prohibited network rules updates. 

It contains a script that generates a set of scenarios where always one network rule is updated to an unacceptable value. This scenario can be then executed normally. 

As a convenient script, one more script is added to run all generated scenarios, printing their results at the end.


rpl-ffl (approved):
Thank you! Will add this to the baseline tests.

PR #50, opened 2025-03-31
removes unused nodes list

kjezek:
This PR simplifies code of `NodeLogParser`. It removes the internal list of `nodes` registered in this parser, as it was not used for any meaningful purpose.


rpl-ffl (approved):
Thank you!

PR #53, opened 2025-04-03
Add logger flags to norma command line tools.

LuisPH3:
This PR adds command line flags to do basic configuration of the logger. This is a quality of life improvement to be able to use debug log level, scope logs to modules, or to reduce verbosity.

Default behavior is equivalent to the previous status quo. 
Flag is global, affects all Norma subcommands.


kjezek (changes requested):
Do I understand correctly that this change configures `go-ethereum` logger? If so, I think we should not use it in Norma. I grep the code and it indeed turns out that `go-ethereum` logger is used at a few places, but I think it was just a miss-import, and we should not rely on a logger library from another project (except of course, using a dedicated logging library)

LuisPH3:
> Do I understand correctly that this change configures `go-ethereum` logger? If so, I think we should not use it in Norma. I grep the code and it indeed turns out that `go-ethereum` logger is used at a few places, but I think it was just a miss-import, and we should not rely on a logger library from another project (except of course, using a dedicated logging library)

This is not entirely correct, norma uses slog in a number of places. Traces and debug calls as well, which currently cannot be printed because of the logger default settings.
The code introduced just installs a custom configuration for slog handler. We could try to do so ourselves by creating a slog handler object, but because this is a debug tool it hardly justifies the effort.

Could you try it? you will see how it starts printing information that before was unreachable.

driver/norma/main.go
```diff
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"os"
 
+	globalflags "github.com/0xsoniclabs/norma/driver/global_flags"
```
kjezek, line 23: would not it be easier to name the directory `globalflags` the same way the package is called?

kjezek (dismissed):
thanks for explanation, I see now that this code also configures the logger triggered from the Norma dependency, i.e. `go-ethereum` types, not only Norma logging itself.

driver/norma/main.go
LuisPH3: I am struggling a little bit with go naming convention, when names get complicated you need multiple words. I usually see folder names having two words, while copilot suggested this naming for the module usage.

I will change it as you describe, but I would like to know how do you address these situations? Do you look for a single word package name?

LuisPH3: changed in: https://github.com/0xsoniclabs/norma/pull/53/commits/991c79d0ca77faabf13631336a50177c30ae1397

kjezek: I try to use single word package names, but when necessary I use multiple words without a separator.

driver/globalflags/logger.go
HerbertJordan, line 1: Some documentation would be nice.

kjezek (dismissed):
thx

driver/norma/main.go
```diff
@@ -40,6 +41,13 @@ func main() {
 			&diffCommand,
 		},
 	}
+
+	app.Flags = append(app.Flags, globalflags.AllLoggerFlags...)
+
+	app.Before = func(c *cli.Context) error {
+		return globalflags.SetupLogger(c)
+	}
```
HerbertJordan, line 49: nit:
```suggestion
	app.Before = globalflags.SetupLogger
```

driver/norma/main.go
```diff
@@ -40,6 +41,13 @@ func main() {
 			&diffCommand,
 		},
 	}
+
+	app.Flags = append(app.Flags, globalflags.AllLoggerFlags...)
+
+	app.Before = func(c *cli.Context) error {
```
HerbertJordan, line 47: nit: both of these could be set up directly in the struct initialization above, right?

HerbertJordan (dismissed):
Nice quality of life improvement, thanks!

HerbertJordan (approved):
Nice!

PR #56, opened 2025-04-10
introduces scenario to update min base fee

kjezek:
this PR introduces a new scenario that updates `minBaseFee`.

It is supposed to test if this network rule does not cause network split when applied to old and a new client

### DIFF
--- a/scenarios/test/min_base_fee_update.yml
+++ b/scenarios/test/min_base_fee_update.yml
@@ -1,0 +1,52 @@
+# This scenario tests update of base fee.
+# It tries this update on a network with
+# pre-Allegro and Allegro nodes.
+# Since this update process has been changed,
+# the network has to split.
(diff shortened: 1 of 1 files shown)


scenarios/test/min_base_fee_update.yml
```diff
@@ -0,0 +1,52 @@
...
+
+# The name of the scenario
+name: Base Fee Network Split
+
+# The duration of the scenario's runtime, in seconds.
+duration: 180
+
+# Initial validator nodes in the network.
+validators:
+  - name: validator-v2.0.4
```
rpl-ffl, line 17:
```suggestion
  - name: validator-v2.0.3
```

rpl-ffl (changes requested):
Not sure if this should be named 2.0.3 to match the `imagename`?

rpl-ffl (approved):

PR #60, opened 2025-06-05
Enable v2.0.4

rpl-ffl:
- Add client v2.0.4 to the scenarios
- Modify scenarios that involves multiple versions such that it includes each available 2.0.x versions

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ BUILD_DIR := $(CURDIR)/build
 .PHONY: all test clean
 
 # Define a list of client versions
-CLIENT_VERSIONS := v2.0.3 v2.0.2 v2.0.1 v2.0.0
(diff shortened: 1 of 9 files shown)


kjezek (commented):
I would not add all client version in all scenarios. It produces a lot of nodes running, and it may become tricky to run these scenarios on a decent hardware.

If we ever need to run scenarios combining all clients, I guess it would be a task of release testing and dedicated scenarios should be created for this.

The scenarios in `scenarios/test` should be more basic

kjezek:
> @kjezek do you find `scenario/release` a more reasonable location?

Is not there any repository where release testing scripts are maintained? Perhaps it should go there? It is not harm to have them in this repository, but I am not sure it will be ever used.

PR #64, opened 2025-06-09
Separate checksum test into 2 tests

rpl-ffl:
Currently, in `local_test.go` there is a test that checks if `LocalNetwork_Can_Run_Multiple_Client_Images`.
It has a list of images
https://github.com/0xsoniclabs/norma/blob/4d52b2d2da11fb46f23ba326656b58846fde1aac/driver/network/local/local_test.go#L490
and uses a set to check if there is a corresponding checksum (so we should get 5 checksums for the versions above).

During testing, there are times when: 
- `sonic:latest` and `sonic:local` is the same
- `sonic:latest` is the same as one of the predetermined versions such as `sonic:v2.0.2`

In such cases, the set will return only 4 checksums and result in timeout.

This PR allows separate the test into two:
- `_LatestVersion` checks for `sonic` and `sonic:local`.
- `_TaggedVersion` checks for `sonic:v2.0.3`, `sonic:v2.0.2`, `sonic:v2.0.1`, `sonic:v2.0.0` - and the checksums must be unique.


kjezek (approved):
looks good, thanks

kjezek (changes requested):
actually, I was wrong with my previous approval. This change would not work.

If all images returned the same checksum, this test would pass. It would not detect what this test is suppose to detect.

rpl-ffl:
@kjezek Make sense. I find it useful that there's a test to check that `sonic:latest` and `sonic:local` exist. This can be its own separate, more simple test.

Will work on it and check in.

kjezek (changes requested):
Thanks for this update. I think it is a good idea to have two separate tests. 

But I think the change has an issue that it duplicates the code that is 99% the same.  I think it should be extracted to a shared method

rpl-ffl:
> Thanks for this update. I think it is a good idea to have two separate tests.
> 
> But I think the change has an issue that it duplicates the code that is 99% the same. I think it should be extracted to a shared method

Agreed. Will get working on the suggestion.

driver/network/local/local_test.go
```diff
@@ -488,48 +490,71 @@ func TestLocalNetwork_Can_Run_Multiple_Client_Images(t *testing.T) {
...
-				line := scanner.Text()
-				if strings.Contains(line, "Sonic binary checksum:") {
-					parts := strings.SplitN(line, ":", 2)
-					if len(parts) == 2 {
-						checksum <- strings.TrimSpace(parts[1])
-					}
-				}
-			}
-		}()
+	if err := net.Shutdown(); err != nil {
```
kjezek, line 519: this looks like redundant, because `net.Shutdown` is already registered `in t.Cleanup`

kjezek (dismissed):
Looks good, thanks

kjezek (approved):
thanks

PR #67, opened 2025-06-10
Add AdvanceEpoch to network rule

rpl-ffl:
Add AdvanceEpoch to rules.go

### DIFF
--- a/genesis/network/rules.go
+++ b/genesis/network/rules.go
@@ -50,7 +50,7 @@ func ApplyNetworkRules(backend ContractBackend, rules genesis.NetworkRules) erro
 	return nil
 }
 
-// AdvanceEpoch trigger the sealing of an epoch and advance the epoch by the given number.
+// AdvanceEpoch triggers the sealing of an epoch advancing it by the given number.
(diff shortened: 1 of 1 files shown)


genesis/network/rules.go
```diff
@@ -49,3 +49,37 @@ func ApplyNetworkRules(backend ContractBackend, rules genesis.NetworkRules) erro
 
 	return nil
 }
+
+// AdvanceEpoch trigger the sealing of an epoch and advance the epoch by the given number.
```
kjezek, line 53:
```suggestion
// AdvanceEpoch triggers the sealing of an epoch advancing it by the given number.
```

kjezek (approved):
Looks good, thanks.

When the feature is completed, it would be good to have also an integration test that would advance the epoch on a running node, and checks that the epoch was moved.

rpl-ffl:
>When the feature is completed, it would be good to have also an integration test that would advance the epoch on a running node, and checks that the epoch was moved.

Note, thanks.

kjezek (approved):
looks good, thansk

PR #68, opened 2025-06-10
Network can now AdvanceEpoch

rpl-ffl:
By calling the genesis/network `AdvanceEpoch`, a network can now advance an epoch.


kjezek (approved):
looks good, thanks

PR #72, opened 2025-06-11
Add sonic_single_proposal scenario

rpl-ffl:
Add a scenario that
- Start with Sonic 
  - enable single-proposer protocol -> new client can join
  - disable single-proposer protocol -> new client can join.


kjezek (approved):
looks good, thanks

PR #73, opened 2025-06-11
Add allegro_single_proposal scenario

rpl-ffl:
Add a scenario that
- Start with Allegro
  - enable single-proposer protocol -> new client can join
  - disable single-proposer protocol -> new client can join.


kjezek (approved):
I think image definitions may be removed, `sonic` is the default

kjezek (approved):
thx

PR #74, opened 2025-06-13
Add slack to blockHeightChecker, defaults to 5 blocks

rpl-ffl:
In blockHeightChecker, currently we allow slack of 1
https://github.com/0xsoniclabs/norma/blob/dcece938304f4d94f86dbfbb0385d5fdcaf4cfd9/driver/checking/block_height.go#L68

There are scenarios where the client would have caught up if given 1-2 seconds, failing the scenario even though this should be tolerable.

This PR add an additional configuration to blockHeightChecker called `slack`, set to 5. This allows us to set the slack for the scenarios above.

Currently the configuration is not accessible from scenario, and if necessary will be added in future PR.

### DIFF
--- a/driver/checking/block_height.go
+++ b/driver/checking/block_height.go
@@ -18,29 +18,27 @@ package checking
 
 import (
 	"fmt"
+	"github.com/0xsoniclabs/norma/driver"
+	"github.com/0xsoniclabs/norma/driver/monitoring"
 	"maps"
 	"strconv"
 	"strings"
-
-	"github.com/0xsoniclabs/norma/driver"
-	"github.com/0xsoniclabs/norma/driver/monitoring"
 )
 
 // allow block height to fall short by this amount
 // slack of 5 means that block 95-99 is also accepted when max block height = 100
-const defaultSlack = int64(5)
+const defaultSlack = 5
 
 func init() {
 	RegisterNetworkCheck("block_height", func(net driver.Network, monitor *monitoring.Monitor) Checker {
-		var slack = defaultSlack
-		return &blockHeightChecker{net: net, slack: &slack}
+		return &blockHeightChecker{net: net, slack: defaultSlack}
 	})
 }
 
 // blockHeightChecker is a Checker checking if all Opera nodes achieved the same block height.
 type blockHeightChecker struct {
 	net   driver.Network
-	slack *int64 // optional slack
+	slack uint8
 }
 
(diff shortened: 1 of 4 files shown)


kjezek (changes requested):
It looks good. Thanks. A minor remark. I think allowing for `slack` being optional makes the code unnecessarily complicated. If no slack is needed, it can be just configured to equal to zero - why sending `nil`, assuming it means zero, if zero can be set directly.  

I was originally thinking of more (maybe over) complicated approach - querying repeatedly blocks that fall behind until they reach the last block. It would make 100% sure that they really were only delayed, and did not crash just a few blocks before the end. But maybe this simple approach can be good enough.

rpl-ffl:
I made it optional to account for both:
https://github.com/0xsoniclabs/norma/blob/2f4b60713760b468efc96877158a8e9c650a47ef/driver/checking/block_height_test.go#L44
https://github.com/0xsoniclabs/norma/blob/2f4b60713760b468efc96877158a8e9c650a47ef/driver/checking/block_height_test.go#L84

But thinking more about it - you're right. We can just make this non-optional.

driver/checking/block_height.go
```diff
@@ -25,15 +25,20 @@ import (
 	"strings"
 )
 
+// allow block height to fall short by this amount
+// slack of 5 means that block 95-99 is also accepted when max block height = 100
+const defaultSlack = uint8(5)
```
kjezek, line 30: I guess casting to uint8 is redundant, right?

kjezek (dismissed):
looks good, thanks

driver/checking/block_height.go
rpl-ffl: I made slack `uint8` because there's no good position to check if the value assigned is < 0.

https://github.com/0xsoniclabs/norma/blob/ec522bae26bf1e5dcb05a5ef4f03a999564174b1/driver/checking/block_height.go#L41

Thinking about it more, now I see I could introduce an intermediate newBlockHeightChecker function and make the check there. (Replacing the following line)

https://github.com/0xsoniclabs/norma/blob/ec522bae26bf1e5dcb05a5ef4f03a999564174b1/driver/checking/block_height.go#L34

kjezek: I ment just if this syntax would work:

```
const defaultSlack = 5
```

kjezek (approved):
thanks, looks good

PR #76, opened 2025-06-17
Move all checker inits to run.go

rpl-ffl:
We now have a need to configure checkers in scenario:

The 3 checkers are fixed.

This PR starts a series of PR to allow the scenario configuration to enable and parameterize checkers as necessary.

This PR focuses on moving the initialization of each checkers from `func init()` in each checkers to `driver/norma/run.go`. 

Incoming PRs will:
- add to scenario configuration a `checker` configurator.
- modify `driver/norma/run.go` so that it defaults to the current 3 checkers but deviate when checkers are specified in scenario.
- convert checkers's `Check()` to `Check(time float32)` so that checkers know to ignore check signal outside its timing window.
- Add new checkers as necessary.

### DIFF
--- a/driver/checking/block_height.go
+++ b/driver/checking/block_height.go
@@ -25,14 +25,10 @@ import (
 	"strings"
 )
 
-// allow block height to fall short by this amount
-// slack of 5 means that block 95-99 is also accepted when max block height = 100
-const defaultSlack = 5
-
-func init() {
-	RegisterNetworkCheck("block_height", func(net driver.Network, monitor *monitoring.Monitor) Checker {
-		return &blockHeightChecker{net: net, slack: defaultSlack}
-	})
+func NewBlockHeightChecker(slack uint8) Factory {
+	return func(net driver.Network, monitor *monitoring.Monitor) Checker {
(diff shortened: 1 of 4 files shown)


kjezek (changes requested):
The original design of this code was based on the [open/close principle ](https://medium.com/@ashutoshdev16/open-closed-principle-in-golang-43d576e473d8). 

In particular, it registers objects that are capable of checking certain attributes of the network. They are available throughout the lifecycle of the program. At some point, they can be triggered - it is happening at the end of the scenario at the moment. 

As I understand, the idea is to modify WHEN the check is happening. This should be done by extending implementation in `checker.go`, the idea is:

[This](https://github.com/0xsoniclabs/norma/blob/main/driver/checking/checker.go#L39) may became a map:  `Checks map[string]Checker `

Them it is easy to implement a method:

```
func (c Checks) Check(name string) error {
	return c[name].Check()
}

```

furthermore, as you suggest to enrich this method with a parameter to decide when to execute the check, I do not think it is right. There should be an event triggered by norma, that would call `Check` at demand (which we already have), but at a requested time instead of the end of the execution

PR #78, opened 2025-06-17
Add default checker behavior

rpl-ffl:
This PR do the following:
- Add a separate registration layer so one can register supported type of checks.
- using `init()`, the 3 default behaviors are now registered as before. This is done in after each check types have been registered.
- The factory can also take on a `map[string]string` so that the yml configuration can be configured generically. Example:
```
checkers:
  - name: block_rolling
    type: block_rolling_checker
    config:                              ### this can be parsed and passed on to the factory
      tolerance: 10        
```
- This feels overly complicated, I will have to go through this again when I'm more fresh. Any feedback is welcome.

Note: This PR still uses the old name checkers. This is to be clean up and a new name of checks should be adopted.

Incoming PRs will:

- Implement a way to repeatedly send `Check` signal to checker by name between `Start` and `End`.
- Add new checkers as necessary.

### DIFF
--- a/driver/checking/block_height.go
+++ b/driver/checking/block_height.go
@@ -25,6 +25,10 @@ import (
 	"strings"
 )
 
+// allow block height to fall short by this amount
+// slack of 5 means that block 95-99 is also accepted when max block height = 100
+const defaultSlack = 5
+
 func NewBlockHeightChecker(slack uint8) Factory {
 	return func(net driver.Network, monitor *monitoring.Monitor) Checker {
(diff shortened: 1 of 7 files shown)


kjezek:
Perhaps do the following: Update the implementation so that the checks can be triggered at a configured time. At the moment it is triggered always at the end of the scenario run. Only update the code to trigger at the configured time, only a configured checker.  Think of other complications, i.e. how to pass through extra configurations later.

PR #80, opened 2025-06-19
Parser can parse checks

rpl-ffl:
This PR add to scenario configuration a `checks` configurator.

There are incoming PRs that will:
- add supported `checks` from checking package and add a parser check to raise error if a scenario requests unsupported `check`.
- make executor aware of `checks` and schedule them as simulation events.

----

This PR is a start of a new set of PRs that replace https://github.com/0xsoniclabs/norma/pull/76

Context:
We now have a need to configure checks in scenarios:

The existing checks above are to be **left alone and functioning exactly as before**.

### DIFF
--- a/driver/parser/check_test.go
+++ b/driver/parser/check_test.go
@@ -591,58 +591,61 @@ func TestScenario_AdvanceEpoch_Failure(t *testing.T) {
 }
 
 func TestScenario_Checks_Success(t *testing.T) {
-	scenario := Scenario{
-		Name:     "Test",
-		Duration: 60,
-		Checks: []Check{
-			{Time: 30, Check: "test"},
+	scenarios := []Scenario{
+		{
+			Name:     "Test_Check_Success1",
+			Duration: 60,
+			Checks: []Check{
+				{Time: 30, Check: "test"},
+			},
 		},
-	}
-	err := scenario.Check()
-	if err != nil {
-		t.Errorf("unexpected error: %v", err)
+		{
+			Name:     "Test_Check_Succuss2",
+			Duration: 60,
+			Checks: []Check{
+				{Time: 30, Check: "test"},
+				{Time: 30, Check: "test2"},
+				{Time: 20, Check: "test3"},
+				{Time: 40, Check: "test4"},
+				{Time: 45, Check: "test"},
+			},
+		},
 	}
 
-	scenario2 := Scenario{
-		Name:     "Test",
-		Duration: 60,
-		Checks: []Check{
-			{Time: 30, Check: "test"},
-			{Time: 30, Check: "test2"},
-			{Time: 20, Check: "test3"},
-			{Time: 40, Check: "test4"},
-			{Time: 45, Check: "test"},
-		},
-	}
-	err = scenario2.Check()
-	if err != nil {
-		t.Errorf("unexpected error: %v", err)
+	for _, scenario := range scenarios {
+		t.Run(scenario.Name, func(t *testing.T) {
+			t.Parallel()
+			if err := scenario.Check(); err != nil {
+				t.Errorf("unexpected error: %v", err)
+			}
+		})
 	}
 }
 
 func TestScenario_Checks_Failure(t *testing.T) {
-	scenario := Scenario{
-		Name:     "Test",
-		Duration: 60,
-		Checks: []Check{
-			{Time: -1, Check: "test"},
+	scenarios := []Scenario{
+		{
+			Name:     "Test_Check_Failure1_BeforeSim",
+			Duration: 60,
+			Checks: []Check{
+				{Time: -1, Check: "test"},
+			},
+		},
+		{
+			Name:     "Test_Check_Failure_AfterSim",
+			Duration: 60,
+			Checks: []Check{
+				{Time: 70, Check: "test"},
+			},
 		},
 	}
-	err := scenario.Check()
-	if err == nil || !strings.Contains(err.Error(), "invalid timing for check") {
-		t.Errorf("invalid timing for check was not detected")
-	}
 
-	scenario2 := Scenario{
-		Name:     "Test",
-		Duration: 60,
-		Checks: []Check{
-			{Time: 70, Check: "test"},
-		},
+	for _, scenario := range scenarios {
+		t.Run(scenario.Name, func(t *testing.T) {
+			t.Parallel()
+			if err := scenario.Check(); err == nil || !strings.Contains(err.Error(), "invalid timing for check") {
+				t.Errorf("invalid timing for check was not detected")
+			}
+		})
 	}
-	err = scenario2.Check()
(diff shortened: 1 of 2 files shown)


kjezek (dismissed):
I think `Time` should be `int`, otherwise it looks good.

driver/parser/parser.go
```diff
@@ -62,6 +63,12 @@ type AdvanceEpoch struct {
 	Epochs *int `yaml:",omitempty"` // nil is interpreted as 1
 }
 
+// Check defines what timing to perform what check
+type Check struct {
+	Time  float32
```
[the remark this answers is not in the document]
rpl-ffl: Right now all the timing variable `Time` has type `float32`.

https://github.com/0xsoniclabs/norma/blob/6528a91002db08211fc2a9a3d56b251e9d0636c9/driver/parser/parser.go#L32
https://github.com/0xsoniclabs/norma/blob/6528a91002db08211fc2a9a3d56b251e9d0636c9/driver/parser/parser.go#L62
https://github.com/0xsoniclabs/norma/blob/6528a91002db08211fc2a9a3d56b251e9d0636c9/driver/parser/parser.go#L74

I agree with you, but I imagine this could be its own PR.

driver/parser/check_test.go
```diff
@@ -589,3 +589,60 @@ func TestScenario_AdvanceEpoch_Failure(t *testing.T) {
...
+			{Time: 30, Check: "test"},
+			{Time: 30, Check: "test2"},
+			{Time: 20, Check: "test3"},
+			{Time: 40, Check: "test4"},
+			{Time: 45, Check: "test"},
+		},
+	}
+	err = scenario2.Check()
+	if err != nil {
+		t.Errorf("Check valid but this error occured: %v", err)
```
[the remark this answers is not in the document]
rpl-ffl: Suggestion committed

driver/parser/check_test.go
```diff
@@ -589,3 +589,60 @@ func TestScenario_AdvanceEpoch_Failure(t *testing.T) {
...
+	scenario := Scenario{
+		Name:     "Test",
+		Duration: 60,
+		Checks: []Check{
+			{Time: 30, Check: "test"},
+		},
+	}
+	err := scenario.Check()
+	if err != nil {
+		t.Errorf("Check valid but this error occured: %v", err)
```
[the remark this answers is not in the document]
kjezek: It looks that this is not resolved. 


The reason for this change is following:  I prefer when the messages are brief and contain only useful information - hence here: `unexpected error `-- it is strong and sharp.  

The original message is long, but it contains no extra useful information that the reader could use to debug the error. 
Academically pedantic: it says "Check valid ..." - but it does not make a lot of sense, because there WAS an error, i.e. something was wrong, maybe the check was valid, maybe not. I.e.  I prefer to avoid assumption from the messages, we should avoid misleading the reader.

driver/parser/check_test.go
```diff
@@ -589,3 +589,60 @@ func TestScenario_AdvanceEpoch_Failure(t *testing.T) {
 		t.Errorf("backward epoch advancement was not detected")
 	}
 }
+
+func TestScenario_Checks_Success(t *testing.T) {
```
kjezek, line 593: can you run `scenario` and `scenario2` in a loop? It repeats exactly the same code twice at the moment

driver/parser/check_test.go
```diff
@@ -589,3 +589,60 @@ func TestScenario_AdvanceEpoch_Failure(t *testing.T) {
...
+	scenario := Scenario{
+		Name:     "Test",
+		Duration: 60,
+		Checks: []Check{
+			{Time: 30, Check: "test"},
+		},
+	}
+	err := scenario.Check()
```
kjezek, line 601: can be written inlined 
`if err := scenario.Check(); err != nil { ... }`

driver/parser/check_test.go
```diff
@@ -589,3 +589,60 @@ func TestScenario_AdvanceEpoch_Failure(t *testing.T) {
...
+		Duration: 60,
+		Checks: []Check{
+			{Time: 30, Check: "test"},
+			{Time: 30, Check: "test2"},
+			{Time: 20, Check: "test3"},
+			{Time: 40, Check: "test4"},
+			{Time: 45, Check: "test"},
+		},
+	}
+	err = scenario2.Check()
```
kjezek, line 617: can be inlined as well

driver/parser/check_test.go
```diff
@@ -589,3 +589,60 @@ func TestScenario_AdvanceEpoch_Failure(t *testing.T) {
...
+	if err != nil {
+		t.Errorf("unexpected error: %v", err)
+	}
+}
+
+func TestScenario_Checks_Failure(t *testing.T) {
```
kjezek, line 623: can you run `scenario` and `scenario2` in a loop? It repeats exactly the same code twice at the moment

driver/parser/check_test.go
```diff
@@ -589,3 +589,60 @@ func TestScenario_AdvanceEpoch_Failure(t *testing.T) {
...
+	scenario := Scenario{
+		Name:     "Test",
+		Duration: 60,
+		Checks: []Check{
+			{Time: -1, Check: "test"},
+		},
+	}
+	err := scenario.Check()
```
kjezek, line 631: can be inlined

driver/parser/check_test.go
```diff
@@ -589,3 +589,60 @@ func TestScenario_AdvanceEpoch_Failure(t *testing.T) {
...
+	}
+
+	scenario2 := Scenario{
+		Name:     "Test",
+		Duration: 60,
+		Checks: []Check{
+			{Time: 70, Check: "test"},
+		},
+	}
+	err = scenario2.Check()
```
kjezek, line 643: can be inlined

driver/parser/parser_test.go
```diff
@@ -189,6 +189,22 @@ func TestParseExampleWithAdvanceEpoch(t *testing.T) {
...
+
+checks:
+  - time: 25
+    check: block_rolling
+  - time: 50 
+    check: block_rolling
+`
+
+func TestParseExampleWithChecks(t *testing.T) {
+	_, err := ParseBytes([]byte(withChecks))
```
kjezek, line 202: can be inlined

kjezek (changes requested):
looks good thanks, but I suggest to run asserts on scenarios in a loop. At the moment both tests repeat the same code twice.

driver/parser/check_test.go
rpl-ffl: made inline

driver/parser/check_test.go
rpl-ffl: `scenario` and `scenario2` now in a loop

kjezek (approved):
looks good, thanks

PR #82, opened 2025-06-20
Executor can add checks

rpl-ffl:
This PR make executor aware of `checks` and schedule them as simulation events.

There are incoming PRs that will:
- make executor aware of `checks` and schedule them as simulation events.

----

Context:
We now have a need to configure checks in scenarios:

The existing checks above are to be **left alone and functioning exactly as before**.


rpl-ffl:
Deprecated by https://github.com/0xsoniclabs/norma/pull/84

PR #83, opened 2025-06-20
Add EIP-7702 testing SmartAccount app

thaarok:
This add new app "smartaccount" producing EIP-7702 SetCode transactions.

The application:
* Deploys trivial SmartAccount implementation contract
* The app "user" represents a bundler, who sets codes for 3 of his users in each SetCode transaction
* The transaction also invokes one of the smart account to increment a counter contract from it (the smart account is a caller of the counter) - this suppose to be a trivial simulation of the ERC-4337 bundler tx.

Note: the app reuse existing bundler user, but we need to avoid using the same user too often - the sonic memory pool does not allow multiple pending transaction for a single sender with set code - triggers ErrInflightTxLimitReached error when inserting another tx into the pool. This PR use a circular-registry of 1000 users to avoid it.


LuisPH3 (commented):
Regarding SetCode operations:
 The PR introduces a workload generator using setcode operations eip7702 transactions interacting with a Smart account contract RFC-4337.
 Operation seems correct. some notes:
 - all txs generated by the system are SetCodeTxs, theoretically delegation could be installed once and reused via other transaction types. that would explore more interactions of the 7702 code paths. 
 
 The code seems clean and correct, please address other comments

load/app/accounts_circular.go
```diff
@@ -0,0 +1,42 @@
+package app
+
+import (
+	"fmt"
+	"github.com/0xsoniclabs/norma/driver/rpc"
+)
+
+// AccountsCircularPool is a circular pool of accounts.
+// Allows reusing a single set of accounts without reusing the same account too shortly after its last usage.
```
kjezek, line 9:
```suggestion
// It allows for reusing a single set of accounts without reusing the same account too shortly after its last usage.
```

kjezek (approved):
looks good, thanks

kjezek (approved):

PR #85, opened 2025-06-20
implements idea of triggering checks based on scenario schedule

kjezek:
This PR shows, how registered checkers can be triggered via an event from scenario confiuration


rpl-ffl (approved):
Thank you for the base PR

PR #87, opened 2025-06-24
Improve advance epoch test

rpl-ffl:
This PR improves advance epoch test by:
- using inline error checking idioms
- remove redundant code by loop over scenarios


kjezek (approved):
looks much better, thanks

PR #88, opened 2025-06-25
Mock Checker

rpl-ffl:
This PR extends https://github.com/0xsoniclabs/norma/pull/85 so that it passes the test.

- Initialize checkers in `InitNetworkChecks`
- Split `TestExecutor_TestUserAbort` into another test `TestExecutor_RunScenarioWithDefaultChecks`
  - `TestExecutor_TestUserAbort` now only tests user abort without any checkers
  - `TestExecutor_RunScenarioWithDefaultChecks` now replaces actual implementation of checkers with mocks and test that all registered checkers are called.

### DIFF
--- a/driver/checking/checker.go
+++ b/driver/checking/checker.go
@@ -47,7 +47,7 @@ func RegisterNetworkCheck(name string, factory Factory) {
 
 // InitNetworkChecks initializes the Checks with the given network.
 func InitNetworkChecks(network driver.Network, monitor *monitoring.Monitor) Checks {
-	var checkers map[string]Checker = make(map[string]Checker, len(registrations))
+	checkers := make(map[string]Checker, len(registrations))
 	for name, factory := range registrations {
 		checker := factory(network, monitor)
 		checkers[name] = checker
--- a/driver/executor/executor_test.go
+++ b/driver/executor/executor_test.go
@@ -256,9 +256,9 @@ func TestExecutor_RunScenarioWithDefaultChecks(t *testing.T) {
 		return checkBlocksRolling
 	})
 
-	checkBlockHeight.EXPECT().Check().Return(nil).Times(1)
-	checkBlocksHashes.EXPECT().Check().Return(nil).Times(1)
-	checkBlocksRolling.EXPECT().Check().Return(nil).Times(1)
+	checkBlockHeight.EXPECT().Check().Return(nil)
+	checkBlocksHashes.EXPECT().Check().Return(nil)
+	checkBlocksRolling.EXPECT().Check().Return(nil)
 
 	checks := checking.InitNetworkChecks(net, nil)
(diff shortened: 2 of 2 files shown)


driver/checking/checker.go
```diff
@@ -45,7 +47,7 @@ func RegisterNetworkCheck(name string, factory Factory) {
 
 // InitNetworkChecks initializes the Checks with the given network.
 func InitNetworkChecks(network driver.Network, monitor *monitoring.Monitor) Checks {
-	var checkers map[string]Checker
+	var checkers map[string]Checker = make(map[string]Checker, len(registrations))
```
kjezek, line 50:
```suggestion
	checkers := make(map[string]Checker, len(registrations))
```

driver/executor/executor_test.go
```diff
@@ -222,8 +223,7 @@ func TestExecutor_TestUserAbort(t *testing.T) {
 		syscall.Kill(syscall.Getpid(), syscall.SIGINT)
 	}).Return(node, nil)
 
-	checks := checking.InitNetworkChecks(net, nil)
-	if err := Run(clock, net, &scenario, checks); err == nil {
+	if err := Run(clock, net, &scenario, nil); err == nil {
```
kjezek, line 226: I do not understand what does this change bring

driver/executor/executor_test.go
```diff
@@ -232,6 +232,44 @@ func TestExecutor_TestUserAbort(t *testing.T) {
...
+		return checkBlocksHashes
+	})
+	checkBlocksRolling := checking.NewMockChecker(ctrl)
+	checking.RegisterNetworkCheck("blocks_rolling", func(driver.Network, *monitoring.Monitor) checking.Checker {
+		return checkBlocksRolling
+	})
+
+	checkBlockHeight.EXPECT().Check().Return(nil).Times(1)
+	checkBlocksHashes.EXPECT().Check().Return(nil).Times(1)
+	checkBlocksRolling.EXPECT().Check().Return(nil).Times(1)
```
kjezek, line 261:
```suggestion
	checkBlockHeight.EXPECT().Check().Return(nil)
	checkBlocksHashes.EXPECT().Check().Return(nil)
	checkBlocksRolling.EXPECT().Check().Return(nil)
```

driver/parser/parser_test.go
```diff
@@ -203,22 +203,6 @@ func TestParseExampleWithChecks(t *testing.T) {
...
-var withChecks = smallExample + `
-
-checks:
-  - time: 25
-    check: block_rolling
-  - time: 50 
-    check: block_rolling
-`
-
-func TestParseExampleWithChecks(t *testing.T) {
```
kjezek, line 215: why this test is deleted?

kjezek (approved):
looks good, thanks. Just there are a few changes, which I do not understand why the were made. In particular, deleted tests.

PR #89, opened 2025-06-30
Add a memory limit to Sonic instances

HerbertJordan:
This PR adds a soft memory limit to `sonicd` instances of 1 GiB per instance. This should help making resources requirements of clients more predictable.


rpl-ffl (approved):
Thank you!

PR #90, opened 2025-06-30
Move network tools from genesis module to the Norma module

HerbertJordan:
The network package seems to have been accidentally placed in the `genesis` module instead of the `norma` module. This PR fixes this issue.

Beyond the code organisational issue, this also caused VS-Code to fail in code navigation. In the long run, it may be considered to eliminate the unusual construct of a nested module in this repository -- namely the `genesis` package.


rpl-ffl (approved):
Thank you. Please let include @kjezek since he also works with this part of the code as well.

PR #91, opened 2025-06-30
Make epoch advancement block until new epoch has been reached

HerbertJordan:
This PR updates the advance-epoch operation to block until the required epoch change has actually happened on the network.

The changes comprise the following steps:
- moved the `AdvanceEpoch` function from the `rules.go` file into its own `epochs.go` file
- extended `AdvanceEpoch` to wait for the network to have reached the new epoch
- added a log message after the epoch change listing the active validators for the new epoch

### DIFF
--- a/driver/network/backend.go
+++ b/driver/network/backend.go
@@ -18,10 +18,11 @@ package network
 
 import (
 	"encoding/hex"
+	"strings"
+
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	"github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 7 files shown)


rpl-ffl (approved):
Thank you!

HerbertJordan:
@rpl-ffl I had to remove the 10 minute to 10 second change for waiting for transaction time-outs since some tests depended on the long time-outs.

Right now, every operation is using the same time-out. We should consider refactoring the system to allow call-specific timeouts for finer granularity in adjusting maximum wait delays.

PR #94, opened 2025-07-01
Check can now be configured

rpl-ffl:
This PR extends Check such that they can be configured as such: `check.Configure(map[string]string).Check()`
- Parser now allows configuration of checks as follow:
```
checks:
  - time: 25
    check: block_rolling
  - time: 30
    check: block_rolling
    config:                        # this will be pass onto Configure
       tolerance: 5
```
- Checker Interface now requires `Configure(map[string]string) (Checker, error)` which, when necessary, will deep copy the original checker and configure the copied instance. The reason we need to deep copy is because we need multiple instances of the same type of checker.
- Each existing implementation of checker has its `Configure` function implemented and tested.
- Executor now `Configure` each check before scheduling it.

### DIFF
--- a/driver/checking/checker.go
+++ b/driver/checking/checker.go
@@ -35,6 +35,7 @@ var registrations = make(registry)
 // Checker does the consistency check at the end of the scenario.
 type Checker interface {
 	Check() error
+	Configure(map[string]string) (Checker, error)
 }
 
 // Checks is a slice of Checker.
--- a/driver/checking/block_height.go
+++ b/driver/checking/block_height.go
@@ -41,6 +41,31 @@ type blockHeightChecker struct {
 	slack uint8
 }
 
+// Configure returns a deep copy of the original checker.
+// If the config doesn't provide any replacement value, copy from the value of the original.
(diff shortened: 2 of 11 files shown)


driver/checking/checker.go
```diff
@@ -35,6 +35,7 @@ var registrations = make(registry)
 // Checker does the consistency check at the end of the scenario.
 type Checker interface {
 	Check() error
+	Configure(map[string]string) (Checker, error)
```
kjezek, line 38: It would be interesting to define a new type like:  `type CheckerConfig map[string]any`, and the method would be then
```
Configure(CheckerConfig) (Checker, error)
```

it would be semantically richer, and also would allow for passing complex json structures in it, if every needed in the future

kjezek (approved):
It looks good, thanks. I put in an idea of even a bit better way of configuring the checker

driver/checking/checker.go
rpl-ffl: Agreed, will follow up on this in a separate PR.
Thank you!

PR #95, opened 2025-07-02
Introduce CheckerConfig

rpl-ffl:
This PR introduces `type CheckerConfig map[string]any` to replace `map[string]string`, mainly to be consumed by `Check.Configure(config CheckerConfig)`.


kjezek (approved):
looks good, thanks

PR #96, opened 2025-07-02
Add check for block gas rate

rpl-ffl:
This PR introduces a new Check `block_gas_rate`. It has a configurable `ceiling`, defaults to `30M`. 
At the end of the scenario, it checks that for each block, the gas rate does not exceed the ceiling.

This helps check the stress test scenario and ensure that gas rate doesn't exceed the intended number in the scenario.


kjezek (approved):
looks good thanks.

I think the header of this file is obsolete, but I am not sure other files were already updated, We can do it in a separate PR

PR #98, opened 2025-07-02
Refactor Monitoring Data

rpl-ffl:
This PR refactors the interface `checking.MonitoringData` to be its own file.
Currently, this is part of `blocks_rolling.go` as it is only being used by the `blocks_rolling` check.
Given that another check `block_gas_rate` is now also using the interface, it makes sense to extract it from `blocks_rolling.go`.


kjezek (dismissed):
thanks for improving the codebase

PR #99, opened 2025-07-03
Improve logging in event processing loop

HerbertJordan:
This change introduces some slight improvements in the logging output to improve the developer experience when debugging issues.

In particular, log messages for the completion of execution steps are added. Furthermore, delayed event starts or terminations are reported as a warning.


rpl-ffl (approved):
Thank you!

PR #100, opened 2025-07-03
Add support for shutting down initial validators

HerbertJordan:
This PR introduces the support for a `end` field the configuration of validators used during network start-ups.

When configuring an initial validator to get offline at a given time, it is undelegating its stake and shutting down analog to what a dynamically introduced validator would do.

This extension facilitates the simulation of scenarios where the set of validators is fully exchanged over time.


rpl-ffl (approved):
Thank you!

PR #104, opened 2025-07-09
Validator ends without unregistration

rpl-ffl:
This PR introduces a way to configure disgraceful exit of nodes.
This is meant to simulate a node's unintentional exit, ultimately to create blackout scenario where >33% of validators suddenly exit.

- Parser allows node to be configured with the timing `kill`, default to `nil`. When set, that network will call `KillNode` on that node.
```
node:
  - name: test
    start: 10
    kill: 20
```
- `network` interface now also includes `KillNode`.
- Executor understands `kill`
   - If both `end` and `kill` is configured, it will fail the scenario `Check()`.

### DIFF
--- a/driver/parser/check.go
+++ b/driver/parser/check.go
@@ -157,7 +157,7 @@ func (n *Node) Check(scenario *Scenario) error {
 	}
 
 	if n.End != nil && n.Kill != nil {
-		errs = append(errs, fmt.Errorf("node cannot both have end and kill; end=%f, kill=%f", *n.End, *n.Kill))
+		errs = append(errs, fmt.Errorf("node cannot have both end and kill; end=%f, kill=%f", *n.End, *n.Kill))
 	}
 
 	if err := checkTimeInterval(n.Start, n.End, scenario.Duration); err != nil {


driver/executor/executor_test.go
```diff
@@ -126,6 +127,77 @@ func TestExecutor_RunMultipleNodeScenario(t *testing.T) {
...
+		node2.EXPECT().Cleanup(),
+	)
+
+	if err := Run(clock, net, &scenario, nil); err != nil {
+		t.Errorf("failed to run scenario: %v", err)
+	}
+	want := Seconds(10)
+	if got := clock.Now(); got < want {
+		t.Errorf("scenario execution did not complete all steps, expected end time %v, got %v", want, got)
+	}
```
kjezek, line 198: I think this is unrelated test

kjezek (commented):
I am not sure what would make the code more complicated if there was a dedicated `kill` syntax. It looks to me that it would just trigger another event that would differ from the `end` event. 

Also I would expect that abrupt shutdown does not impact only unstacking, but really kills the node without properly closing DB, etc.  I know that it is not needed now, but could be useful in the future for example to test recovery.

Since we already have `Kill` command on a `Node` , I would think it is much straighforward to reuse it. 

If there is something I am missing, we can merge this PR, but first I would like to understand why just calling `Kill` from an event associated to `kill` syntax would not do the job

rpl-ffl:
@kjezek you're absolutely right. Please find the updated version of this PR that uses kill.

driver/executor/executor.go
```diff
@@ -311,11 +315,23 @@ func scheduleNodeEvents(node *parser.Node, queue *eventQueue, net driver.Network
...
+		killTime = Seconds(*node.Kill)
+	}
 	endTime := end
 	if node.End != nil {
 		endTime = Seconds(*node.End)
 	}
 
+	// if killed before ended, no longer need to signal end
+	// kill takes precedence over end, if somehow configured at the same time
+	if killTime <= endTime {
```
kjezek, line 329: perhaps easier would be to prevent running scenario where both `kill` and `end` is configured, as part of `Check`. It could be even more user friendly, because a user can be confused what is happening when unintentionally configures both.

driver/executor/executor.go
```diff
@@ -349,34 +365,50 @@ func scheduleNodeEvents(node *parser.Node, queue *eventQueue, net driver.Network
 			},
 		))
 
-		queue.add(toSingleEvent(
-			endTime,
-			fmt.Sprintf("[%s] Stop Node", name),
-			func() error {
-				if instance == nil {
+		if killTime != NoKill {
```
kjezek, line 368: I think condition `node.Kill != nil` would be more readable, it looks to me that the constant `NoKill` just make the code more complicated without clear benefit

driver/executor/executor.go
```diff
@@ -349,34 +365,50 @@ func scheduleNodeEvents(node *parser.Node, queue *eventQueue, net driver.Network
...
+			queue.add(toSingleEvent(
+				killTime,
+				fmt.Sprintf("[%s] Kill Node", name),
+				func() error {
+					if instance == nil {
+						return nil
+					}
+					if err := net.KillNode(*instance); err != nil {
+						return err
+					}
```
kjezek, line 378: should not we also call:

```
if err := net.RemoveNode(*instance); err != nil {
					return err
				}
				if err := (*instance).Stop(); err != nil {
					return err
				}
				if err := (*instance).Cleanup(); err != nil {
					return err
				}
```
?

there are listeners in network that may need to get information that the node is gone

driver/executor/executor.go
```diff
@@ -349,34 +365,50 @@ func scheduleNodeEvents(node *parser.Node, queue *eventQueue, net driver.Network
...
-					return err
-				}
-				return nil
-			},
-		))
+				},
+			))
+		}
+
+		if endTime != NoEnd {
```
kjezek, line 384: the same here, it looks to me that the newly introduced constant `NoEnd` only makes the code more complicted

kjezek (changes requested):
Hi, thanks for this change. Generally it looks good, I added a few comments where I think the code should be improved. 

Functionality wise: I think even when the node is killed, it should be properly removed. Otherwise I think there will be dangling containers in docker. 

Code clean-up wise: new constants `NoKill` and `NoEnd` seem redundant to me. They are just assigned based on configuration, and further program flow is based on them, but it can be based on the original configuration I guess

Finally, when user configures both `end` and `kill`, I think the most straight forward  way is to treat this as misconfiguration

driver/executor/executor.go
rpl-ffl: Removed `NoEnd` and uses `node.Kill == nil` directly

driver/executor/executor.go
```diff
@@ -349,34 +349,59 @@ func scheduleNodeEvents(node *parser.Node, queue *eventQueue, net driver.Network
...
-					return err
-				}
-				return nil
-			},
-		))
+				},
+			))
+		}
+
+		if node.Kill == nil {
```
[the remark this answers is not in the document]
rpl-ffl: `node.Kill == nil` -> `node.End != nil`

kjezek (approved):
looks good thanks. 

Please try to avoid force push when asking for a repeated review, it makes it harder to check what has been changed from the last review round.

PR #106, opened 2025-07-15
App controller checks if network has no node

rpl-ffl:
This PR modifies `load/controller`'s `GetReceivedTransactions()` method so that it checks if the network is empty.
If so, it returns an error that the `periodic_source` knows to catch and do nothing and not panic.

This solves the problem that currently, there's a timing between last node removal and monitoring shutdown that a periodic source could `ReadValue()`. When such a call happen, the monitor shutdown will fail, sometimes panic.

### DIFF
--- a/driver/monitoring/utils/periodic_source.go
+++ b/driver/monitoring/utils/periodic_source.go
@@ -19,7 +19,6 @@ package utils
 import (
 	"errors"
 	"math/rand"
-	"strings"
 	"time"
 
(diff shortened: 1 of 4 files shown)


rpl-ffl:
Both suggestions make a lot of sense. Will take note of them moving forward.
Thank you!

HerbertJordan (dismissed):

HerbertJordan (approved):

PR #110, opened 2025-07-21
Support disabling default checks

rpl-ffl:
This PR introduces a way to disable default checks.

Currently we have 4 checks that are enabled by default:
- `block_height` - pass iff the maximum difference between node's block height is less than some slack (defaults to 5)
- `blocks_hashes` - pass iff each node provides the same hashes for all blocks/stateRoots.
- `blocks_rolling` - pass iff for any time interval (defaults to 10 samples) there is at least one node with block height increase.
- `block_gas_rate` - pass iff block gas rate does not exceed a ceiling (defaults to `math.MaxFloat64`)

In blackout scenario, we need to turn off `blocks_rolling` since during blackout, no block height increase is expected for all nodes.
Thus, we need a way to disable default checks.

### DIFF
--- a/driver/executor/executor.go
+++ b/driver/executor/executor.go
@@ -65,8 +65,36 @@ func run(
 		return nil
 	}))
 
+	// schedule custom network checks
+	// since this makes a clone of registered checks, it must be done before
+	// deletion of disabled checks.
+	for _, c := range scenario.Checks.CustomChecks {
+		checker := checks.GetCheckerByName(c.Check)
+		if checker == nil {
+			return fmt.Errorf("custom check type '%s' not found", c.Check)
+		}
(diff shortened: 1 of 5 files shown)


kjezek (changes requested):
Hi, thanks for this update. 

I am a bit sceptical, if the new configuration is not too complicated. 

What if we simplify it as follows:
1. When the `checks` block is **not** defined -- all checks are executed
2. When the `checks` block **is** defined -- it switches to a custom mode, and the user has to define all checks he wants, no default checks are triggered at all.

The idea is that whenever the user updates this configuration, the scenario very likely targets some very specific situations, and that it is up to the user to configure what he wants to check.

PR #111, opened 2025-07-21
Custom checks disable default checks

rpl-ffl:
This PR introduces a way to disable default checks.

Whenever custom checks are specified in a scenario, it disables default check and it is up to the user to define exactly the checks they need.
- An example of default check is now provided in `scenario/test/custom_checks.yml`
- This replaces https://github.com/0xsoniclabs/norma/pull/110.


kjezek (approved):
this is much more elegant. thanks for update.

PR #112, opened 2025-07-22
Add error checker

rpl-ffl:
This PR adds 2 functionalities to checkers: 
- `failingChecker` which is a `checker` that expects an error. This can be set through the config as `failing: true`
- `blocksRollingChecker` now accepts additional configuration `start`, defaults to 0. If set, it will take the interval `[start, latest)` when checking if blocks are rolling.

This allows the blackout scenario to express checks as so:
```
checks:

  # this checks that block height increases from 0 to 100
  - time: 100
    type: blocks_rolling
    
  # this checks that block height does _not_ increase from 100 to 200
  - time: 200
    type: blocks_rolling
    config:
      failing: true                      # this expects error
      start: 100                         # this set starting position timing 100
      
  # this checks that block height increases from 200 to 300
  - time: 300
    type: blocks_rolling
    config:
      start: 200                         # this set starting position timing 200
```


driver/executor/executor.go
```diff
@@ -85,6 +85,24 @@ func run(
 					return fmt.Errorf("error configuring checks; %v", err)
 				}
 
+				if c.Check == "blocks_rolling" {
```
kjezek, line 88: I am not trying fully understand what this condition is exactly doing, but it seems very wrong to have a specific condition for one of the checkers. All the logic should be inside the checker

scenarios/test/blackout.yml
```diff
@@ -34,20 +32,40 @@ nodes:
...
+    check: block_height
+  - time: 199
+    check: blocks_hashes
+  # custom checks for blackout
+  - time: 80
+    check: blocks_rolling
+  - time: 140
+    check: blocks_rolling
+    config:
+      error: network is down
```
kjezek, line 54: I do not understand how the user is supposed to know, which error message should be used here. We should not expect the user studies the code of Norma. Since the error message is out-of-control of the user, I think we should only define `error : true`, maybe better `failing : true` to be consistent with other configurations we already have

driver/checking/block_gas_rate.go
```diff
@@ -50,6 +50,10 @@ func (c *blockGasRateChecker) Configure(config CheckerConfig) (Checker, error) {
 		return c, nil
 	}
 
+	if _, exist := config["error"]; exist {
```
kjezek, line 53: We should detect that we expect an error, and wrap the checker into this checking one somewhere here:
https://github.com/0xsoniclabs/norma/blob/main/driver/executor/executor.go#L83

Otherwise, each checker is enriched with this condition, which leads to code repetition, and is error prone because when a new checker is created, this condition will be easily forgotten.

kjezek (changes requested):
Hi, thanks for this contribution. I did just first review round, I can immediately see two points of improvements:
* We should detect that the error is expected in the `executor` and wrap the checker into the error one there, it is not a good code style to repeat the condition in every checker
* expecting a user to define, which error message produced is not practical, as the user cannot know what error message is produced

driver/executor/executor.go
```diff
@@ -80,11 +80,33 @@ func run(
...
+						return checking.NewErrorChecker(checker, c.Config)
+					}
+				}
+
 				configured, err := checker.Configure(c.Config)
 				if err != nil {
 					return fmt.Errorf("error configuring checks; %v", err)
 				}
 
+				if val, exist := c.Config["start"]; exist {
```
kjezek, line 98: this is exactly the same logic, which I disputed before. The only difference is that a new kind of checker `NewMonitorChecker` is used, which has inside it the same logic which was before here.  This is not really what we want. It breaks the original design. 

The original idea is:
* Checker is registered via names
* The checker is picked-up  here by name
* The checker has `Configure` method, which is called
* All the customisation of the checker is done only inside `Configure` method
* This method returns a new instance of the `Checker`, which is configured

From this, there is no reason to create any new copies, or new configurations at this point in the code.  All should happen in `Configure`. I

driver/executor/executor.go
```diff
@@ -80,11 +80,33 @@ func run(
 					return fmt.Errorf("check '%s' not found", c.Check)
 				}
 
+				if val, exist := c.Config["failing"]; exist {
```
kjezek, line 83: We should wrap all checker in `NewErrorChecker`, and only apply its functionality when `failing` is configured. It would streamline the code here

driver/checking/blocks_rolling.go
```diff
@@ -13,12 +13,18 @@ func init() {
 	RegisterNetworkCheck("blocks_rolling", func(net driver.Network, monitor *monitoring.Monitor) Checker {
 		return &blocksRollingChecker{monitor: &monitoringDataAdapter{monitor}, toleranceSamples: defaultToleranceSamples}
 	})
+	RegisterNetworkCheck("blocks_rolling_position", func(_ driver.Network, monitor *monitoring.Monitor) Checker {
```
kjezek, line 16: I think it is an overkill to have an extra checker, which is only wrapping `blocks_rolling`. All customisation can be done in `Configure` method, which was introduced for this reason,

driver/checking/checker.go
```diff
@@ -75,3 +88,45 @@ func (c Checks) Check() error {
...
+
+func (c *failingChecker) Check() error {
+	if err := c.checker.Check(); err == nil {
+		return fmt.Errorf("failing expected")
+	}
+	return nil
+}
+
+func (c *failingChecker) Configure(config CheckerConfig) (Checker, error) {
+	checker, err := c.checker.Configure(config.copyExceptError())
```
kjezek, line 115: I think that method `copyExceptError` makes the code unnecessarily complicated. What is the reason to remove something from the configuration. If the wrapped checker does not support a field, it will just ignore it. It means there is not a lot of reasons to remove something from the configuration

driver/checking/checker.go
```diff
@@ -75,3 +88,45 @@ func (c Checks) Check() error {
...
+	if err != nil {
+		return nil, err
+	}
+
+	return &failingChecker{checker: checker}, nil
+}
+
+// NewMonitorChecker returns a checker that extract informations from monitor
+// Implementation of what to check/how to update is up to each individual checker
+func NewMonitorChecker(checker Checker) (Checker, error) {
```
kjezek, line 125: This is bad, for every new configuration, someone would have to go here, and update the code here. All configurations should be captured in each `Checker`

driver/checking/blocks_rolling.go
```diff
@@ -90,3 +102,46 @@ func (c *blocksRollingChecker) Check() error {
...
+	return nil
+}
+
+// Configure sets the target blocksRollingChecker
+func (c *blocksRollingPositionChecker) Configure(config CheckerConfig) (Checker, error) {
+	if config == nil {
+		return c, nil
+	}
+
+	val, exist := config["target"]
```
kjezek, line 136: who sets `target` ?

driver/checking/blocks_rolling.go
kjezek: furthermore, it seems that this registration is not used at all, it is bypassed by this line:
```
monChecker, err := checking.NewMonitorChecker(checker)
```

kjezek (changes requested):
Hi, I went through the code, there is a lot of design flaws which I commented on inline. 

Overall, If I understand goal of this PR, it may be done very straightforward updating the code we already had before.

* scenario defines `start` and `time` time (which is end time)  for the `blocksRollingChecker`
* This checker now checks block from `zero` to `time` (last bar of the time series)

I think only a tiny modification is needed:
* `start` time is passed to the checker via `Configure`
* when the checker is triggered, it does not check from `zero`, but from `start` instead
* that is it, 

Regarding the `ErrorChecker` it is almost ok, but maybe I would wrap all checker in this one, just the logic if error is expected or not would be handled inside, not in the `executor`, to keep the `executor` code as simple as possible

driver/checking/checker.go
rpl-ffl: Removed [4b74f75](https://github.com/0xsoniclabs/norma/pull/112/commits/4b74f75c4764cd1efd0c38314c1afefcb0fe6c67)

driver/checking/blocks_rolling.go
rpl-ffl: removed [eda53c5](https://github.com/0xsoniclabs/norma/pull/112/commits/eda53c5aa643a27df672c45ec126ee6040ac662f)

driver/executor/executor.go
rpl-ffl: Reworked as suggested [06a7367](https://github.com/0xsoniclabs/norma/pull/112/commits/06a7367ecb2da1195469701f82bfd356d4b1149c)

driver/executor/executor.go
rpl-ffl: finalized here: [06a7367](https://github.com/0xsoniclabs/norma/pull/112/commits/06a7367ecb2da1195469701f82bfd356d4b1149c)

driver/checking/checker.go
```diff
@@ -75,3 +77,42 @@ func (c Checks) Check() error {
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to configure checker; %v", err)
+	}
+
+	val, exist := config["failing"]
+	if !exist {
+		return configured, nil
+	}
+
+	failing, ok := val.(bool)
```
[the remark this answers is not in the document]
rpl-ffl: reworked check data type into parsing [9cb1fab](https://github.com/0xsoniclabs/norma/pull/112/commits/9cb1fab5b45d299024d5e2b84ed1acb5db855254)

kjezek (changes requested):
thanks for this update. It looks much better, just picking up data from the TimeSeries is s till not correct. You are still matching position in the Series with time, but the time is the value, not the position. 

Furthermore, at places, parsing integer from string would be probably simpler, than checking if the value is int, float, etc.

rpl-ffl:
@kjezek 
To apply `strconv.Atoi`, to my eyes, it requires that `CheckerConfig` be changed from `map[string]any` to `map[string]string`. Otherwise, the any interface is already associated with underlying type during parsing so it is already either a `bool` or an `int`.

If we do change `CheckerConfig` from `map[string]any` to `map[string]string`, there will be a lot of trivial changes such as 
`CheckerConfig{"ceiling": 50}` -> `CheckerConfig{"ceiling": "50"}` I just check in with you first if you agree with this direction before carrying this out.

driver/checking/blocks_rolling.go
```diff
@@ -62,12 +64,28 @@ func (c *blocksRollingChecker) Check() error {
...
 			nodeFunctional = false //node produced no blocks
 			continue
 		}
-		items := series.GetRange(0, last.Position)
+
+		var first monitoring.Time = 0
+		found := false
+		dataPoints := series.GetRange(0, last.Position+1)
+		for _, dp := range dataPoints {
+			if dp.Position == monitoring.Time(c.start) {
```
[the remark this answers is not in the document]
rpl-ffl: @kjezek 
Regarding point 2: I'm not following. Here is my understanding:

In this line https://github.com/0xsoniclabs/norma/blob/d0d1e8e10fda6cfe47b42d36afd5e4b68f7b631b/driver/checking/blocks_rolling.go#L76
the `monitoring.Series[monitoring.Time, monitoring.BlockStatus].GetRange` would return `[]DataPoint[monitoring.Time, monitoring.BlockStatus]` and `dp` as an element in the slice would have this behavior:
- `dp.Position` should return `monitoring.Time`
- `dp.Value` should return `monitoring.BlockStatus`

I couldn't see why we would use dp.Value in this case.

driver/checking/blocks_rolling.go
```diff
@@ -62,12 +64,28 @@ func (c *blocksRollingChecker) Check() error {
...
+				first = dp.Position
+				found = true
+				break
+			}
+		}
+		if !found {
```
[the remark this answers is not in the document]
rpl-ffl: No longer `return err` but `return nil`.
https://github.com/0xsoniclabs/norma/pull/112/commits/65b8ec1d22f625155017b9e8f0ef0fdd0e91124c

driver/checking/blocks_rolling.go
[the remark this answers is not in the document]
kjezek: sorry, my bad. You are right, `Position` is `Time` for `TimeSeries`.

driver/parser/check.go
```diff
@@ -115,9 +115,9 @@ func (s *Scenario) Check() error {
 		}
 	}
 
-	for _, chk := range s.Checks {
-		if chk.Time < 0 || chk.Time > s.Duration {
-			errs = append(errs, fmt.Errorf("invalid timing for checks: %f", chk.Time))
+	for _, c := range s.Checks {
```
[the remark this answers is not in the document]
rpl-ffl: removed pollution here [faf8bc7](https://github.com/0xsoniclabs/norma/pull/112/commits/faf8bc721b3fccf2db69c872dbaca78ac1bb5e70)

driver/checking/checker.go
```diff
@@ -75,3 +74,55 @@ func (c Checks) Check() error {
...
+	configured := c.checker.Configure(config)
+	if failing, exist := config["failing"]; !exist || !failing.(bool) {
+		return configured
+	}
+	return &failingChecker{configured}
+}
+
+type CheckerConfig map[string]any
+
+func (config *CheckerConfig) Check() error {
```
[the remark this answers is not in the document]
rpl-ffl: - All variants of int now no longer support; only int is supported.

Covered by compromise outlined in https://github.com/0xsoniclabs/norma/pull/112#issuecomment-3131238936

[9a63ed8](https://github.com/0xsoniclabs/norma/pull/112/commits/9a63ed8940a4b0637c8652e9fcba50499df9ca09)

driver/checking/checker.go
```diff
@@ -75,3 +74,55 @@ func (c Checks) Check() error {
...
+func (c *failingChecker) Check() error {
+	if err := c.checker.Check(); err == nil {
+		return fmt.Errorf("failing expected")
+	}
+	return nil
+}
+
+func (c *failingChecker) Configure(config CheckerConfig) Checker {
+	configured := c.checker.Configure(config)
+	if failing, exist := config["failing"]; !exist || !failing.(bool) {
```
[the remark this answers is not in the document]
rpl-ffl: No changes, but covered by compromise outlined in https://github.com/0xsoniclabs/norma/pull/112#issuecomment-3131238936

[9a63ed8](https://github.com/0xsoniclabs/norma/pull/112/commits/9a63ed8940a4b0637c8652e9fcba50499df9ca09)

driver/checking/checker.go
```diff
@@ -75,3 +74,53 @@ func (c Checks) Check() error {
...
+
+type CheckerConfig map[string]any
+
+func (config *CheckerConfig) Check() error {
+	errs := []error{}
+
+	isBool := func(v any) bool { _, ok := v.(bool); return ok }
+	isPositiveInt := func(v any) bool { i, ok := v.(int); return ok && i >= 0 }
+
+	checks := map[string]func(any) bool{
```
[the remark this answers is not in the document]
rpl-ffl: Added as a future issue here: https://github.com/0xsoniclabs/norma/issues/114

driver/checking/blocks_rolling.go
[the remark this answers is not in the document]
rpl-ffl: > I think this may be misleading, because when the user miss configures this Check, it will behave like everything run smoothly.

> I know that it may sound that I am controlling myself, but I was not precise enough in my message above - I wanted to say that the whole if condition should be removed, it would then check from zero.

> Also returning an error here would make sense, but skipping a test when it is misconfigured may lead to confusions.

Sorry, I'm a bit confused. Please let me clarify.
This is the behavior I intended:
- If say we have `datapoints := [1, 2, 3, 4, 5]` and the user asks for `start=10`, then the logic would set `found=false`, return an error and fail the scenario.
- This, as you pointed out, assumes that the user correctly set `start=10` so there must be data at this point. If the user had misconfigured, it's not very forgiving.

This is the behavior I think you mean:
- If say we have `datapoints := [1, 2, 3, 4, 5]` and the user asks for `start=10`, then we should just continue as though `start=0`.

Is this right?

kjezek (approved):

PR #117, opened 2025-09-25
Update norma genesis generation to support gas subsidies feature

LuisPH3:
Sonic's gas subsidies feature relies on a global registry of subsidies installed as a smart contract at a fix address. 

This PR installs the current contract code at the required address if the scenario sets the `UPGRADES_GAS_SUBSIDIES` environment variable to `true`

Testing has been added, and one test has been extended to test different upgrades.

### DIFF
--- a/genesis/app/genesis_test.go
+++ b/genesis/app/genesis_test.go
@@ -4,12 +4,13 @@ import (
 	"encoding/json"
 	"flag"
 	"fmt"
-	"github.com/0xsoniclabs/sonic/integration/makefakegenesis"
-	"github.com/urfave/cli/v2"
 	"os"
 	"path"
 	"strings"
 	"testing"
+
+	"github.com/0xsoniclabs/sonic/integration/makefakegenesis"
+	"github.com/urfave/cli/v2"
 )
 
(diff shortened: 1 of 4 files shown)


LuisPH3:
> lgtm, ty!
> 
> A quick question from me: I see that we have added an additional test
> 
> https://github.com/0xsoniclabs/norma/blob/6b68c12e75c6937512afd018b01f2ee68c1a1172/genesis/app/genesis_test.go#L84-L136
> 
> rather than modifying the original test to also check for `UPGRADES_GAS_SUBSIDIES`.
> 
> https://github.com/0xsoniclabs/norma/blob/6b68c12e75c6937512afd018b01f2ee68c1a1172/genesis/app/genesis_test.go#L16-L82
> 
> Is there any benefit in creating an additional test rather than modifying existing ones? No strong feeling one way or the other from me, but this pattern will come up a lot so I just want to check now.

The reason why I decided to duplicate the test is because the existing test sets a collection of flags, while the new one only checks the existence of a single flag. Because these two are orthogonal to each other (sonic features and allegro features can exist with and without subsidies)  I did not want to pollute existing tests. 
The execution time is fast enough to have an extra function.

PR #119, opened 2025-10-01
Add load app to test gas subsidies

LuisPH3:
GasSubsidies is implemented in form of transactions requesting sponsorship by setting gas price 0.
If the subsidies registry contract has funds allocated for such, the transaction can be executed. 

The load app will allocate a large sum of funds to sponsor the call to the counter app. 
The counter contract is deployed by the app to avoid collision with other instances in the network.  
The counter can be used to verify the number of sponsorships executed. 

An example scenario is provided. 
- The scenario starts a background load of transactions to see the impact of introducing sponsored transactions
- some time later, starts creating sponsored transactions. The effect can be seen on the build-up on the tx_pool.
- before finishing the exercise, generation stops, to observe the pool being able to dispatch queued transactions. 

image

### DIFF
--- a/load/app/app_test.go
+++ b/load/app/app_test.go
@@ -43,7 +43,11 @@ func TestGenerators(t *testing.T) {
 	if err != nil {
 		t.Fatalf("failed to create new local network: %v", err)
 	}
-	t.Cleanup(func() { _ = net.Shutdown() })
+	t.Cleanup(func() {
+		if err := net.Shutdown(); err != nil {
+			t.Fatalf("failed to shutdown network: %v", err)
+		}
+	})
 
 	primaryAccount, err := app.NewAccount(0, PrivateKey, nil, FakeNetworkID)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


load/app/subsidies_app.go
```diff
@@ -0,0 +1,172 @@
...
+		}
+		addresses[i] = workerAccount.address
+	}
+
+	// Fund all accounts with 1 FTM so they can pay for the balance transfer
+	// gas costs shall be covered by the subsidy
+	err := appContext.FundAccounts(addresses, big.NewInt(1e18))
+	if err != nil {
+		return nil, fmt.Errorf("failed to fund accounts; %w", err)
+	}
```
HerbertJordan, line 131: Is this still needed? It should be possible for a 0-balanced account to run sponsored transactions, right?

HerbertJordan (changes requested):
Looks good. 

I would, however, suggest to remove the balance from the user accounts just to make sure that these transactions are not accidentally payed by the users themselves. We should stress the 0-balanced user case.

rpl-ffl (approved):
Thank you!

kjezek (dismissed):

HerbertJordan (dismissed):

HerbertJordan (approved):

PR #123, opened 2025-10-03
Rpc_test now checks Call instead of TransactionReceipt

rpl-ffl:
This PR modifies `driver/rpc/rpc_test.go` so that it checks `Call` instead of `TransactionReceipt`. 
The reason `TransactionReceipt` is no longer used can be found here: https://github.com/0xsoniclabs/norma/pull/121.


HerbertJordan (approved):
Thanks for fixing those tests!

PR #125, opened 2025-10-07
Add new versions to docker images

LuisPH3:
This PR extends the list of images available to Norma with the new release branch and the existing tags in it. 
Building an image from the release branch directly allows testing features from the branch before they get tagged. 

- v2.0 and v2.1 are branches, they will build an image with the HEAD of that branch at the time make is invoked. To be used during experiments.
- Vx.y.z are tags, they represent code released under that tag and will not change between builds. To be used for testing.

The need for adding tags comes from the need to test the current release branch features, which haven't been tagged yet.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,9 @@ BUILD_DIR := $(CURDIR)/build
 .PHONY: all test clean
 
 # Define a list of client versions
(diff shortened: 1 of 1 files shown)


kjezek (approved):
thanks

kjezek (approved):

kjezek (approved):

PR #126, opened 2025-10-07
Use tag sonic:main instead of sonic

LuisPH3:
This is an optional cleanup which makes the tagging of the sonic branch explicit.
This avoids confusion between "sonic" and "sonic:local" tags. 

The reason why this PR is optional is that it may break downstream projects that rely on the existence of the "sonic" tag.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ BUILD_DIR := $(CURDIR)/build
 .PHONY: all test clean
 
 # Define a list of client versions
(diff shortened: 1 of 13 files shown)


kjezek (commented):
I am not sure about this change. I think `latest` or empty tag is idiomatic in Docker, while `main` is idiomatic in git for both to represent the last available version. 

but I understand that maybe for people not familiar with docker may make more sense to use `main` for both

PR #128, opened 2025-10-07
Add more gas subsidies scenarios

LuisPH3:
This PR adds scenarios to test gas subsidies:
- two validators with subsidies feature, but different versions can do gas subsidies
- Validators with versions not having gas subsidies, drop from the network. 

This PR creates a folder for gas subsidies tests and migrates the already existing scenario to it. If this is a problem for downstream projects/tests, this can be reverted.


scenarios/gas_subsidies/compatible.yml
```diff
@@ -0,0 +1,32 @@
...
+
+duration: 5
+
+network_rules:
+  genesis:
+    UPGRADES_GAS_SUBSIDIES: true
+
+validators:
+  - name: main
+    instances: 1
```
kjezek, line 15: one instance is default, i.e. can be omitted

kjezek (approved):
some of the values can be omitted as they are default values. Otherwise it looks good

Issue #124, opened 2025-10-06
rpc.Client cannot be mocked


LuisPH3:
The type `rpc.Client` contains embeddings of third party interfaces. These cannot be mocked, and attempting to regenerate the mocks fails.
This is a problem since has been mocked in the past, and the mock for the type is used in many tests. The mock and the type are no longer in sync.

PR #129, opened 2025-10-07
Fix Mock generation

LuisPH3:
This PR uses mockgen in package mode to be capable of full type introspection. This allows mocking of types with embedded third-party interfaces. 

From `mockgen --help`
> Package mode works by specifying the package and interface names.
It is enabled by passing two non-flag arguments: an import path, and a
comma-separated list of symbols.
You can use "." to refer to the current path's package.
Example:
	mockgen database/sql/driver Conn,Driver
	mockgen . SomeInterface

fixes: https://github.com/0xsoniclabs/norma/issues/124


driver/network/gen.go
```diff
@@ -0,0 +1,22 @@
...
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with Norma. If not, see <http://www.gnu.org/licenses/>.
+
+package network
+
+// The following generator uses mockgen in package mode (opposed to source mode as in other files)
+// to have full reflection of the imported interfaces. This allows mocking embedded interfaces from other packages
+
+//go:generate mockgen -destination network_mock.go -package network . ContractBackend,Host
```
HerbertJordan, line 22: Could we keep this line in the file defining the interface (`backend.go`) instead of introducing a new `gen.go` file?

I quickly tested
```
//go:generate mockgen -destination backend_mock.go -package network . ContractBackend
```
in `backend.go` and it seems to work fine.

This way we could keep our naming schema following the `<something>.go` and `<something>_mock.go` and we do not need to introduce `gen.go` files in those packages.

HerbertJordan (changes requested):

driver/network/gen.go
LuisPH3: yes we can.

kjezek (approved):

PR #147, opened 2026-04-08
Add ecdsa app (Osaka/Brio hardfork)

thaarok:
This PR extends Norma's transaction load infrastructure with a new application type:

* EcdsaCounter — exercises the P256Verify precompile (EIP-7951, address 0x100); a fixed secp256r1 key pair is generated at deployment and each transaction submits a valid P-256 signature over a random hash.


rpl-ffl (dismissed):

simonlechner (approved):
There is an [integration test](https://github.com/0xsoniclabs/sonic/blob/main/tests/secp256r1_precompile_test.go) in sonic that insures that the precompile is enabled with brio and has the correct gas price. This test has more emphasis on the correctness of the verify which I think is a good addition.
LGTM

PR #149, opened 2026-04-09
Add app deploying large contract (Brio size limit increase)

thaarok:
This PR extends Norma's transaction load infrastructure with a new application type - LargeContract:

This tests new contract size limits introduced in Brio:
* The **initialization** code size limit is doubled from 49152 b to 98304 b
* The **runtime** code size limit of contrast is doubled from 24576 b to 49152 b

The test use two large contracts:
* LargeContractCounter: init=52531 b  runtime=48956 b
* LargeContract: init=92529 b  runtime=48904 b

(utility script `contract_sizes.sh` provides the info above)

The test:
* Initialization deploys LargeContractCounter with counter initialized to 0
* Generated transactions deploys LargeContract, which calls LargeContractCounter incrementation as part of the constructor
* Padding functions and useless operations are added into both contracts to make the contracts larger

Side note:
* GasFeeCap decreased from 1e13 to 1e12 (sufficient) to avoid errors because of insufficient balance for (GasFeeCap+GasTip)*GasLimit


scripts/contract_sizes.sh
LuisPH3, line 1: Is this a utility?  Used for development?

thaarok: Yes, this script allows to check the size of LargeContract/LargeContractCounter, to provide you info how many padding code to add - it provides output like the one in the PR:
```
    LargeContractCounter: init=52531 b runtime=48956 b
    LargeContract: init=92529 b runtime=48904 b
```

simonlechner: I think it would be nice to add a unit test that calls the script and checks the contract size. If somebody ever removes the unused functions or a new solidity compiler starts to optimize does away, we would not notice.

simonlechner (dismissed):
It is good that this change is also checked on a Norma level. As discussed, a failing test case pre-Brio should be added as a standalone test file.

load/app/large_contract_app_test.go
LuisPH3, line 1: Unit testing in norma are good to check sanity of what we write but they do not really check anything on a scenario.
The interesting scenario would be to test the situation with a mixture of client versions at both pre-brio and post-brio hardfork.  All nodes must come to the same sane conclusion. 
For this to be tested, the norma load generator needs to be aware if the transaction is accepted and executed or not, and check expectations accordingly.

This unit test does not check mixture of node versions and introduction of the new upgrade, that needs to be defined at the generator level. right?

thaarok: To test this, not a norma unit test, but a norma scenario should be run.
I think you can already define scenario like that and norma will check at the end of the scenario that block hashes matches - no need to test this on the load generators side.

scripts/contract_sizes.sh
thaarok: Covered by new negative pre-Brio test - if the contract compiles smaller, the deployment pre-brio will succeed and the test will fail.

simonlechner (approved):

LuisPH3 (approved):

PR #150, opened 2026-04-10
Add app "mix" mixing all types of transactions

thaarok:
This PR extends Norma's transaction load infrastructure with a new application type - this mixes all types of transactions, to avoid the need to extend all testing scenarios to "flavor" all of them with a small amount of all possible types of transactions everytime when a new type is added.
(The scenario needs to specify the "mix" app to include this mix of transactions.)

Also apps testing have been refactored to be shorter and to include the general factory call.


load/app/mix_app.go
```diff
@@ -0,0 +1,163 @@
...
+// The probability of picking an app equals its weight divided by the sum of
+// all weights.
+type mixEntry struct {
+	appType string
+	weight  int
+}
+
+// mixAppTypes lists the application types included in the mix together with
+// their relative selection weights.
+var mixAppTypes = []mixEntry{
```
LuisPH3, line 38: nit: as a follow-up, we could consider registering apps differently, so that someone implementing a new app cannot forget updating this list.

LuisPH3 (approved):

PR #152, opened 2026-04-13
Run R reporting in docker - remove R dependency

thaarok:
Removes dependency on locally installed R, process reporting in a docker container.

Also modifies the `make norma` target to not rebuild docker images - unnecessary development slowdown.


LuisPH3 (approved):

LuisPH3 (commented):
This PR removes the need for manually installing R in machines using Norma. Norma requires docker access, no new dependency is needed.

PR #154, opened 2026-04-15
Fix DeployContract error handling

thaarok:
Fix of error handling during load generator initialization:

`appContext.DeployContract` or `appContext.Run` helper methods can currently return receipt with failed status but some of their callers check only if they returns err.

This fix modifies DeployContract and Run to return err when the receipt status is failed.
It also removes the receipt status check where redundant after this change.


load/app/erc20_app.go
```diff
@@ -130,15 +130,12 @@ func (f *ERC20Application) CreateUsers(appContext AppContext, numUsers int) ([]U
...
-	receipt, err := appContext.Run(func(opts *bind.TransactOpts) (*types.Transaction, error) {
+	_, err = appContext.Run(func(opts *bind.TransactOpts) (*types.Transaction, error) {
 		return erc20Contract.MintForAll(opts, addresses, big.NewInt(1_000000000000000000))
 	})
 	if err != nil {
 		return nil, fmt.Errorf("failed to mint ERC-20 for all users; %w", err)
 	}
-	if receipt.Status != types.ReceiptStatusSuccessful {
-		return nil, fmt.Errorf("failed to mint ERC-20 for all users; transaction reverted")
-	}
```
LuisPH3, line 141: why is this one deleted ?
won't this be a problem? 
I would expect something to happen, silently dropping errors is like try with empty caches, a pretty bad habit that can bite you back in the future.

thaarok: The check is being added into the `appContext.Run()` so this check is redundant then.

LuisPH3 (approved):

PR #155, opened 2026-04-16
Add bundles load generator (AllOfBundle)

thaarok:
Initial implementation of bundles load generator:
* Only AllOf successful bundles generated for now


load/app/allofbundle_app.go
```diff
@@ -0,0 +1,227 @@
...
+	}
+
+	currentBlock, err := u.client.BlockNumber(context.Background())
+	if err != nil {
+		return nil, fmt.Errorf("failed to get current block number: %w", err)
+	}
+
+	envelope := bundle.NewBuilder().
+		WithSigner(u.signer).
+		SetEnvelopeSenderKey(u.bundler.privateKey).
```
LuisPH3: I am afraid that this will not be sufficient. 
Envelopes are never executed and therefore they will not increment nonce. This code will run into a deadlock. 
For the current version of bundles, it is required that envelopes are sent with a one-off key. This will not be changed before 2.2

LuisPH3: Once the RPC methods are in place, we can use submit-bundle. Which will create this key silently. 
When submitting envelopes as normal txs, this responsibility is on the envelope creator.

LuisPH3 (changes requested):

thaarok:
Updated to use a new sender account for each bundle.
Bundle sender does not need any funding, the envelope tx is sent with zero gasPrice.

LuisPH3 (approved):

LorenzSchueler (approved):

PR #157, opened 2026-04-16
Upgrade dependencies to support bundles

thaarok:
* Dependencies upgraded to be able to use `0xsoniclabs/sonic/gossip/blockproc/bundle` helpers
* Registered UPGRADES_TRANSACTION_BUNDLES in the genesis generator


LuisPH3 (approved):
thanks!

PR #159, opened 2026-04-21
Add SubsidizedBundle app

thaarok:
Adds **SubsidizedBundle** app load generator. This app generates an **AllOf bundle**:
* First tx (by a sponsor) sets sponsorship for an ERC20 allowance setting by an user
* Second tx (by the user) use the sponsorship and sets the allowance
* Third tx (by the sponsor) use the allowance an transfers ERC20 token from the user to sponsor

This simulates a typical defi scenario, where a user wants to invest into a platform and the platform pays the gas for it.


load/app/subsidizedbundle_app.go
```diff
@@ -0,0 +1,278 @@
...
+	}
+
+	currentBlock, err := u.client.BlockNumber(context.Background())
+	if err != nil {
+		return nil, fmt.Errorf("failed to get current block number: %w", err)
+	}
+
+	envelope := bundle.NewBuilder().
+		WithSigner(u.signer).
+		SetEnvelopeSenderKey(bundler.privateKey).
```
LuisPH3: The builder uses one use keys when left unspecified.  you can spare creating an account.

load/app/subsidizedbundle_app.go
```diff
@@ -0,0 +1,278 @@
...
+			}),
+			bundle.Step(u.user.privateKey, &types.DynamicFeeTx{
+				Nonce:     u.user.getNextNonce(),
+				Gas:       approveGasLimit.Uint64(),
+				GasFeeCap: big.NewInt(0), // gasPrice=0: covered by the subsidy from step 1
+				GasTipCap: big.NewInt(0),
+				To:        &u.erc20Address,
+				Data:      approveData,
+			}),
+			bundle.Step(u.sponsor.privateKey, &types.DynamicFeeTx{
```
LuisPH3, line 266: can we have two cases, for successful  and rolled-back bundle?
for example even and odd transaction generations. A rolled back case is always a valid case, this applies to all bundle workloads. 

The reason being that happy cases are not good test candidates, errors happen always in between.

load/app/bundle_app_test.go
```diff
@@ -0,0 +1,124 @@
...
+		envelopeTxs = append(envelopeTxs, tx)
+	}
+
+	// The envelope itself has no receipt - wait for each inner step tx to land on
+	// chain individually, mirroring how testGenerator waits for tx receipts.
+	chainID, err := rpcClient.ChainID(context.Background())
+	if err != nil {
+		t.Fatalf("failed to get chain ID: %v", err)
+	}
+	signer := types.NewCancunSigner(chainID)
```
LuisPH3: nit: use `types.latestSingerForChainId`

load/app/bundle_app_test.go
```diff
@@ -0,0 +1,124 @@
...
+
+	for _, envelopeTx := range envelopeTxs {
+		txBundle, openErr := bundle.OpenEnvelope(signer, envelopeTx)
```
LuisPH3: the correct way of waiting for bundles execution is to use 
`sonic_getBundleInfo` with the plan hash, this can detect rolled-back bundles which otherwise commit no transaction to the block.

LuisPH3 (changes requested):
The generator is correct, but it misses the important case where bundles rolled-back.

driver/rpc/rpc.go
```diff
@@ -45,6 +46,15 @@ type Client interface {
...
 	WaitTransactionReceipt(txHash common.Hash) (*types.Receipt, error)
+
+	// GetBundleInfo calls sonic_getBundleInfo with the given execution plan hash.
+	// Returns nil, nil if the bundle has not been executed yet.
+	GetBundleInfo(planHash common.Hash) (*sonicapi.RPCBundleInfo, error)
+
+	// WaitForBundleInfo polls sonic_getBundleInfo until the bundle has been
+	// executed or the context is cancelled. Returns nil, nil if the context
+	// expires before the bundle is executed.
+	WaitForBundleInfo(ctx context.Context, planHash common.Hash) (*sonicapi.RPCBundleInfo, error)
```
LuisPH3, line 57: 👍

LuisPH3 (dismissed):
The load generator injects both succeeding and rolling-back bundles into the network.

facuMH (approved):

LuisPH3 (approved):

LorenzSchueler (approved):

PR #161, opened 2026-04-23
Add probabilisticaly failing bundle app

thaarok:
Add load generator for bundled transactions randomly failing in cca 10% of cases.
The purpose is to generate transactions which are accepted by the transaction pool, but fails on the chain.

The app generates:
* AllOf bundles of 2 transactions with 5% prob. of revert (10% prob. of failure for the whole bundle)
* OneOf bundles of 2 transactions with 30% prob. of revert (9% prob. of failure for the whole bundle)


LuisPH3 (approved):
Thanks

PR #163, opened 2026-04-24
Handle rejecting failing bundles on sending

thaarok:
* Handle rejecting failing bundles by eth_sendRawTransaction
* Remove random sending of (deterministically) failing bundles
* Improve randomization of randomly failing bundles (include seed from tx params) - avoid mass rejection of multiple txs send from the same sender at the same block heigh


load/app/bundle_app_test.go
```diff
@@ -100,20 +97,21 @@ func testBundleGenerator(t *testing.T, application app.Application, ctxt app.App
...
-			if err != nil {
-				t.Fatalf("failed to open bundle envelope: %v", err)
-			}
-			planHashes = append(planHashes, txBundle.Plan.Hash())
+		txBundle, err := bundle.OpenEnvelope(signer, tx)
+		if err != nil {
+			t.Fatalf("failed to open bundle envelope: %v", err)
 		}
+		planHash := txBundle.Plan.Hash()
+		fmt.Printf("Sent bundle %d (plan %s)\n", i, planHash)
```
LuisPH3, line 105: nit: debug messages, shall we remove them?

LuisPH3 (approved):
This makes sense, deterministically failing bundles will always be dropped at the pool admission. The removed randomization served no purpose in Norma.

load/app/bundle_app_test.go
thaarok: They proved to be useful in these unit tests

PR #165, opened 2026-04-28
Use non-failing nodes only as a random RPC

thaarok:
* Avoid run failures caused by choosing an intentionally failing node in the scenario as a random RPC
* ~~Avoid LargeContract app init failure in the Mix app init when UPGRADES_BRIO is not enabled~~ - removed


load/app/mix_app.go
```diff
@@ -55,7 +55,7 @@ var mixAppTypes = []mixEntry{
 	{"selfdestructoldcontract", 1, nil},
 	{"selfdestructnewcontract", 1, nil},
 	{"ecdsa", 2, nil},
-	{"largecontract", 1, nil},
+	{"largecontract", 1, []string{"UPGRADES_BRIO"}},
```
LuisPH3: Large contract needs to run with older versions and fail/ be rejected. 

this is important because in the presence of a bug, this could create a fork.

thaarok: Fixed: modified mix app to generate all types of txs, regardless of the network config

LuisPH3 (approved):

PR #167, opened 2026-04-28
Modify LargeContract app to require Brio only for txs, not for init

thaarok:
Modify LargeContract app to require Brio only for txs, not for init:
* In the app init, deploy regular Counter contract - does not require Brio
* In generated txs, deploy LargeContracts - each constructor will increment the Counter


load/app/large_contract_app.go
```diff
@@ -43,9 +43,9 @@ func NewLargeContractApplication(ctxt AppContext, feederId, appId uint32) (Appli
 		return nil, fmt.Errorf("failed to get chain ID; %w", err)
 	}
 
-	_, receipt, err := DeployContract(ctxt, contract.DeployLargeContractCounter)
+	_, receipt, err := DeployContract(ctxt, contract.DeployCounter)
```
LuisPH3, line 46: Question, why do we need to deploy a contract at the beginning at all ?  It seems to me like this is an issue resulting from the large contract being a proxy. Why is this needed?

LuisPH3 (approved):
This PR changes the large contract app to not deploy a large contract during app initialization, when execution may fail and invalidate the scenario. Instead, it deploys a smaller contract and the load generated is the deployment of large contracts.

load/app/large_contract_app.go
thaarok: We need a single contract (deployed on init) to count the amount of successful transactions for monitoring.

The issue is not the large contract being a proxy, the issue is the single monitoring Counter contract is large, so it can be deployed only while Brio is enabled.

You are true that there is no need for it. Therefore I change it here to use a regular (small) Counter contract to do this job, while only the generated txs involves a large contract.

thaarok:
Note: The CI is currently failing on issue [#742](https://github.com/0xsoniclabs/sonic-admin/issues/742):
```
12:46:21      --- FAIL: TestGenerators_Bundles/OneOfBundle (6.20s)
12:46:21          bundle_app_test.go:98: bundle is not executable
12:46:21              bundle trial-run failed
```

PR #172, opened 2026-05-04
Fix norma run interrupting using context/signal

thaarok:
When norma run is interrupted during the initialization (waiting for the validator nodes start), norma does not delete created containers.

This PR moves the signal handling from the executor.Run() into its caller main.runScenario(). The executor and the node initialization will get a context which can be cancelled by the interrupt.


LuisPH3 (approved):
Thank you very much. This was a hideous problem. There are scripts to clean the environment, i had to use them so many times.

PR #174, opened 2026-05-05
Add where old nodes drop scenario.

facuMH:
This PR adds a single scenario. In this scenario, a network starts with allegro and brio enabled. 13 sonic 2.2 start alongside 6 nodes of the different sonic 2.1 versions. All the 2.1 should fail. Then it tries to add them again after some time, and they should fail to join the network again.

The aim of this scenario is to verify that older nodes cannot join a network with an unknown configuration to them.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,8 @@ BUILD_DIR := $(CURDIR)/build
 # Define a list of client versions
 CLIENT_VERSIONS := \
 	v2.0 v2.0.0 v2.0.1 v2.0.2 v2.0.3 \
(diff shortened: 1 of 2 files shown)


thaarok:
It seems the scenario is supposed to test, that older client version will stop when they recognize an unknown hardfork, however that is not what "failing: true" do - the scenario will succeed because the **older clients will diverge**.

Two issues I see:
* norma does not allow to enforce a way how is the node supposed to fail
* from what I know, Sonic is not able to detect unsupported configuration - it will not stop, it will diverge

PR #175, opened 2026-05-06
Upgrade golang to 1.26.

facuMH:
This PR updates go version to 1.26.0.
I didn't update the Dockerfile, which refers to 1.26.3 explicitly. We usually don't pin patch version in Dockerfile in Sonic, but I think this is done to avoid mismatches between go and debian image versions

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ Project Norma is a system testing infrastructure for the Sonic blockchain client
 ## Requirements
 
 For building/running the project, the following tools are required:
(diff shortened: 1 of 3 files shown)


thaarok (approved):

thaarok:
@ihtefyw FYI, it is not necessary to upgrade go version of norma, if you only want to test sonic with newer go.
Only the go version in the **Dockerfile** is important and sonic is built using it.
There is go-level dependency of norma on sonic, but this is only because of data-types and bundle utils.
* You need to upgrade the go version in Dockerfile always when you want to test sonic which use newer go.
* You need to upgrade the go version used to build norma only when you want to use newer util for building bundles or other code from sonic in the norma code.

PR #176, opened 2026-05-06
Enable OneOfBundle app test

thaarok:
This enables OneOfBundle app, but lets it rely on obtaining nonces from RPC.

As discussed in https://github.com/0xsoniclabs/sonic-admin/issues/742, we are not going to support sending OneOf bundles in high-rate, it is expected to fail when trying to pass multiple following OneOf bundles.

### DIFF
--- a/load/app/oneofbundle_app.go
+++ b/load/app/oneofbundle_app.go
@@ -171,6 +171,7 @@ type OneOfBundleUser struct {
 }
 
 func (u *OneOfBundleUser) GenerateTx() (*types.Transaction, error) {
+	ctx := context.Background()
 	successfulFirst := rand.Intn(2) == 0
 
 	transferData, err := u.erc20Abi.Pack("transfer", u.targetAddress, big.NewInt(1))
@@ -178,13 +179,22 @@ func (u *OneOfBundleUser) GenerateTx() (*types.Transaction, error) {
 		return nil, fmt.Errorf("failed to pack rich transfer: %w", err)
 	}
 
-	currentBlock, err := u.client.BlockNumber(context.Background())
+	nonceRich, err := u.client.PendingNonceAt(ctx, u.richSender.address)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get nonce for senderA: %w", err)
+	}
+	noncePoor, err := u.client.PendingNonceAt(ctx, u.richSender.address)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get nonce for senderB: %w", err)
+	}
+
+	currentBlock, err := u.client.BlockNumber(ctx)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get current block number: %w", err)
 	}
 
 	successfulStep := bundle.Step(u.richSender.privateKey, &types.DynamicFeeTx{
-		Nonce:     u.richSender.getCurrentNonce(),
+		Nonce:     nonceRich,
 		Gas:       70_000,
 		GasFeeCap: gasFeeCap,
 		GasTipCap: gasTipCap,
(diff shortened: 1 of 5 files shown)


load/app/oneofbundle_app.go
```diff
@@ -171,28 +171,38 @@ type OneOfBundleUser struct {
...
+	ctx := context.Background()
 	successfulFirst := rand.Intn(2) == 0
 
 	transferData, err := u.erc20Abi.Pack("transfer", u.targetAddress, big.NewInt(1))
 	if err != nil {
 		return nil, fmt.Errorf("failed to pack rich transfer: %w", err)
 	}
 
-	currentBlock, err := u.client.BlockNumber(context.Background())
+	nonceRich, err := u.client.PendingNonceAt(ctx, u.richSender.address)
```
LuisPH3: As discussed offline, the pool is not aware of bundled transactions. This workload will run into the same issues as before in some scenarios.
This seems to work at this time because of the cadence of the scenario.
 
This generator will be a source of issues to people unfamiliar with this code if committed as it is. I propose rewriting the workflow so that we do not run into this issue:

We can rewrite the code into a generator which prodices 2 types of bundles with 2 txs:
- oneOf where the first tx succeeds
- oneOf where the first tx fails 

In either of both cases, the succeeding tx can be sent from a normal Norma account (which carries its own nonce counting), nonces are expected to be monotonic.
The other transaction can use a constant nonce as it will never be executed. 

wdyt?

thaarok: You are basically suggesting to return to the state before this PR.

The issue is, if txs will be generated faster than like 1tx/sec, the issue https://github.com/0xsoniclabs/sonic-admin/issues/742 will reappear and a tx will be rejected by txpool.

We have two options:
* use nonce from RPC - this will allow the generator to recover, to pass at least some txs in
* use sequential nonces as you suggest - but when sonic rejects a single tx (because of the issue above), no following tx will be processed (too high nonce - and the missing one will miss forever)

LuisPH3: What I meant is that we should craft the bundles in a way that does not trigger the issue [742](https://github.com/0xsoniclabs/sonic-admin/issues/742), but still being OneOf. 
The issue is triggered because of the involvement of nonces of 2 accounts, lets prevent that and we can use the usual norma nonce generator. 

Does this make sense?

thaarok: True, we can make it working if we use a single sender for both txs - updated.

thaarok:
Updated to use a **single sender** for both txs in OneOfBundle - this workarounds https://github.com/0xsoniclabs/sonic-admin/issues/742.

load/app/oneofbundle_app.go
```diff
@@ -173,52 +161,65 @@ type OneOfBundleUser struct {
...
+		firstStep = bundle.Step(u.sender.privateKey, &types.DynamicFeeTx{
+			Nonce:     u.sender.getNextNonce(),
+			Gas:       70_000,
+			GasFeeCap: gasFeeCap,
+			GasTipCap: gasTipCap,
+			To:        &u.erc20Address,
+			Data:      transferFailingData,
+		})
+		secondStep = bundle.Step(u.sender.privateKey, &types.DynamicFeeTx{
+			Nonce:     u.sender.getNextNonce(), // both will run, both nonces will be consumed
```
LuisPH3, line 209: I am confused here, would the first transaction revert? Then nonce will not be consumed.
This case should work by both transactions having the same nonce.

thaarok: This is the only combination accepted by the sonic client - confirmed by unit tests:
* when first tx in OneOf succeed, it consumes the nonce, but second tx does not run at all, does not consume a nonce
* when first tx in OneOf fail, in consumes the nonce and also the second tx (which succeed) consumes a nonce

LuisPH3: right

LuisPH3 (approved):

PR #185, opened 2026-05-21
Add scenario where validators are upgraded before starting brio

LuisPH3:
The sonic client keeps a local history of bundles executed within the last ~1000 blocks.
This test checks that all nodes start filling this storage at the same block height independently of when the node has been upgraded to 2.2.

This scenario has two validators updated during its execution, one of them is offline during the upgrade.

### DIFF
--- a/genesis/app/validator.go
+++ b/genesis/app/validator.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"crypto/ecdsa"
+	"errors"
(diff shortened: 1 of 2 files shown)


HerbertJordan (dismissed):
Nice, thanks!

LuisPH3:
This PR got blocked, github does not update recent push. 
Ill try to create a new one.

PR #186, opened 2026-05-26
Handle new txpool requirement for using current nonce

thaarok:
Handle new txpool requirement for bundles introduced in https://github.com/0xsoniclabs/sonic/pull/1005.

* wait for previous bundle to be on-chain before sending another one (fix tests)
* load nonces for bundles from rpc (allow recovery when some bundle is rejected by txpool)


LuisPH3 (approved):
The proposed changes use getpendingnonce to estimate the next available nonce for an account. 
This change solves the issue of the client not accepting anymore gapped nonces in bundled transactions.

PR #187, opened 2026-05-28
Add scenario where validators are upgraded before starting brio

LuisPH3:
The scenario runs nodes which upgrade at different times and start using bundles, Epoch hashes must remain coherent.

The sonic client keeps a local history of bundles executed within the last ~1000 blocks.
This test checks that all nodes start filling this storage at the same block height independently of when the node has been upgraded to 2.2.

This scenario has two validators updated during its execution, one of them is offline during the upgrade.

### DIFF
--- a/genesis/app/validator.go
+++ b/genesis/app/validator.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"crypto/ecdsa"
+	"errors"
 	"fmt"
 	"path"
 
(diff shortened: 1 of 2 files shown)


HerbertJordan (approved):
Timing looks sporty -- meaning, there is only little time between events, and in the past stopping and starting nodes took longer than 10 seconds. But this may have changed with recent improvements in client startup time ... so if the test passes, all good.

thaarok (approved):

PR #195, opened 2026-06-08
Add optional NonRoot docker runs

facuMH:
**The Problem**
	Docker by default runs containers as root user. The test [`TestLocalNetwork_MountDataDir_Can_Be_Reused`](https://github.com/0xsoniclabs/norma/blob/main/driver/network/local/local_test.go#L721) runs the following steps which produce the test to fail:

	- The host runner creates a temporary directory owned by the non-root host user.
	- The directory is bind-mounted into the container.
	- The containerized process writes files into this mount point. Because the container runs as root,
	   these files inherit root:root ownership.
	- When the test finishes, the host runner attempts to delete the temporary directory but fails with 
	   a Permission Denied error because it does not have privileges to remove root-owned files.

**The Solution**
	Run the docker using the user id and user group that is currently running the tests so that the files created inside the container are created with the same user that running the tests and can delete the temp folder at the end of the test.

**Implementation Details & Adjustments**
   - Write Permissions in Container Filesystem: Running the image as a non-root user means that all of the file/folder creations that were done in `/` now need to be done somewhere else where the non-root user has writing permission.
   - Network Simulation: Some tests need to simulate network latency, and do so by using the `tc` command. This commands needs to run as root by its very nature, so all tests cannot run with docker as a non-root user. Hence a config option is added to `ContainerConfig`, `NodeConfig` and `operaNodeCondfig`, so that the tests can specify which containers need to run as current user.

### DIFF
--- a/driver/docker/docker.go
+++ b/driver/docker/docker.go
@@ -80,6 +80,7 @@ type ContainerConfig struct {
 	Entrypoint      []string // Entrypoint to run when starting the container. Optional.
 	Network         *Network // Docker network to join, nil to join bridge network
 	DataDirBinding  *string  // mount client datadir to this path on host
+	NonRootUser     bool
 }
 
 // NewClient creates a new client facilitating the creation of Docker
@@ -162,7 +163,8 @@ func (c *Client) Start(config *ContainerConfig) (*Container, error) {
 
 	init := true
 	stopTimeout := int(config.ShutdownTimeout.Seconds())
-	resp, err := c.cli.ContainerCreate(context.Background(), &container.Config{
+
+	containerConfig := container.Config{
 		Image:      config.ImageName,
 		Tty:        false,
 		Env:        envVars,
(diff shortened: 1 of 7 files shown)


driver/docker/docker.go
```diff
@@ -80,6 +80,7 @@ type ContainerConfig struct {
 	Entrypoint      []string // Entrypoint to run when starting the container. Optional.
 	Network         *Network // Docker network to join, nil to join bridge network
 	DataDirBinding  *string  // mount client datadir to this path on host
+	NonRootUser     bool
```
LuisPH3, line 83: missing doc, just like the others.

driver/network/local/local_test.go
```diff
@@ -724,13 +724,11 @@ func TestLocalNetwork_MountDataDir_Can_Be_Reused(t *testing.T) {
 	// jenkins uses different access privileges for docker
 	// i.e. we need to create a temporary directory in /tmp for docker mount
 	// as the test cleanup cannot delete the directory if the mount is in the subdirectory of this test.
-	temp, err := os.MkdirTemp("/tmp", fmt.Sprintf("%s-docker-volume-*", t.Name()))
+	tmp := t.TempDir()
+	temp, err := os.MkdirTemp(tmp, fmt.Sprintf("%s-docker-volume-*", t.Name()))
```
LuisPH3, line 728: this is redundant, t.TempDir is already a new directory.

scripts/run_sonic.sh
```diff
@@ -18,8 +18,7 @@ echo "genesis validator count=${VALIDATORS_COUNT}"
 datadir=$STATE_DB_DATADIR
 # Initialize datadir
 if [[ ! -d "${datadir}/chaindata" ]]; then
```
LuisPH3, line 20: there is something wrong . in go code ` dataDir = "/sonic-client/datadir"` and here `datadir` is added again-

scripts/run_sonic.sh
```diff
@@ -18,8 +18,7 @@ echo "genesis validator count=${VALIDATORS_COUNT}"
...
-  mkdir -p "${datadir}"
```
LuisPH3, line 21: when is the datadir created if not mounted?

LuisPH3 (changes requested):
The changes make sense from the perspective of having to delete the leaked folder. Nevertheless they introduce a new dimension of complexity.
Now there are tests which behave differently if they mount external paths. The flag has been introduced but I do not see any scenario being upgraded to use it. 

I am afraid that this method is too complicated for the problem that we are trying to solve. And it does not even solve it:
- original issue: a temp folder cannot be deleted because it contains root files.
- new status quo after the changes: the test uses t.TempDir and leaks the folder without deleting it because still contains root files. 


Please revert this changes and replace 	`os.MkdirTemp` with `t.TempDir`, that wold do the trick .

facuMH:
This issue will be addressed in a different manner.

PR #197, opened 2026-06-08
Remove parallel from memory hungry tests

LuisPH3:
This PR removes parallel tests spawning containers, although this could yield better execution times than sequential execution, it will most certainly exhaust resources on medium-sized machines. 

My developer machine cannot run Norma unit tests without limits. Execution time is yet within minutes and IMO it does not worth to parallelize this code.


thaarok (dismissed):
ok

thaarok (approved):

PR #199, opened 2026-06-09
Fix reusable mount test.

facuMH:
**The Problem**
Docker by default runs containers as root user. The test [TestLocalNetwork_MountDataDir_Can_Be_Reused](https://github.com/0xsoniclabs/norma/blob/main/driver/network/local/local_test.go#L721) runs the following steps which produce the test to fail:

- The host runner creates a temporary directory owned by the non-root host user.
- The directory is bind-mounted into the container.
- The containerized process writes files into this mount point. Because the container runs as root,
   these files inherit root:root ownership.
- When the test finishes, the host runner attempts to delete the temporary directory but fails with 
   a Permission Denied error because it does not have privileges to remove root-owned files.
   
**The Solution**
Give full permisions (`chmod -r 777`) to the `datadir` folder at the end of the `run_sonic.sh` scrip.
This does not change semantics for any tests and enables any test that wants to do its own cleanup or have a custom datadir to do so, without any extra steps.

### DIFF
--- a/driver/network/local/local_test.go
+++ b/driver/network/local/local_test.go
@@ -724,15 +724,7 @@ func TestLocalNetwork_MountDataDir_Can_Be_Reused(t *testing.T) {
 	// jenkins uses different access privileges for docker
 	// i.e. we need to create a temporary directory in /tmp for docker mount
 	// as the test cleanup cannot delete the directory if the mount is in the subdirectory of this test.
-	temp, err := os.MkdirTemp("/tmp", fmt.Sprintf("%s-docker-volume-*", t.Name()))
-	if err != nil {
-		t.Fatalf("failed to create temporary directory: %v", err)
-	}
-	defer func() {
-		if err := os.RemoveAll(temp); err != nil {
-			t.Fatalf("failed to remove temporary directory: %v", err)
-		}
-	}()
+	temp := t.TempDir()
 
 	config := driver.NetworkConfig{Validators: driver.DefaultValidators, OutputDir: temp}
 	net, err := NewLocalNetwork(t.Context(), &config)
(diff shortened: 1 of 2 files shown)


LuisPH3 (dismissed):

driver/network/local/local_test.go
```diff
@@ -724,13 +724,7 @@ func TestLocalNetwork_MountDataDir_Can_Be_Reused(t *testing.T) {
 	// jenkins uses different access privileges for docker
 	// i.e. we need to create a temporary directory in /tmp for docker mount
 	// as the test cleanup cannot delete the directory if the mount is in the subdirectory of this test.
```
LuisPH3: comment is obsolete

scripts/run_sonic.sh
```diff
@@ -93,3 +93,5 @@ export GOMEMLIMIT="1GiB"
     --datadir.minfreedisk 0 \
     $EXTRA_ARGUMENTS
 
+# datadir folder needs be erasable by any user to enable cleanups.
```
LuisPH3: comment does not mention docker permisions, could be more informative.

driver/network/local/local_test.go
facuMH: https://github.com/0xsoniclabs/norma/pull/199/commits/dc04e95e78b64e20742dff8560c4abb8f428896b

scripts/run_sonic.sh
facuMH: https://github.com/0xsoniclabs/norma/pull/199/commits/4bb7500c0b8f95dec014fe3b5c8bd251c35a0d44

LuisPH3 (approved):
👍

PR #201, opened 2026-06-10
Add linter github action

facuMH:
This PR closes https://github.com/0xsoniclabs/sonic-admin/issues/787 as the last PR in the series of linters.

This PR enables golangci github action to run whenever a PR modifies a `*.go` or `go.mod` file.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -1,0 +1,52 @@
+# Copyright 2024 Fantom Foundation
+# This file is part of Norma System Testing Infrastructure for Sonic.
+#
+# Norma is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Lesser General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Norma is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with Norma. If not, see <http://www.gnu.org/licenses/>.
+
+
+name: Go
+
+permissions:
+  contents: read
+  checks: write
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    paths:
+     - '.github/workflows/go.yaml'
+     - '.golangci.yml'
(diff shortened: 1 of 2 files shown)


.github/workflows/go.yml
```diff
@@ -0,0 +1,52 @@
...
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Norma is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with Norma. If not, see <http://www.gnu.org/licenses/>.
```
LuisPH3: We should not license scripts downloaded from the internet where we modify a line. 
please ignore licenses in .github folder.

.golangci.yml
```diff
@@ -0,0 +1,63 @@
...
+    - staticcheck
+    - unused
+
+  settings:
+    staticcheck:
+      checks:
+        - "all"
+        # Ignored rules
+        - "-ST1000" # Incorrect or missing package comment.
+        - "-ST1003" # Poorly chosen identifier.
```
LuisPH3, line 38: are these really needed? how difficult would it be to solve them?

.golangci.yml
LuisPH3, line 1: to reduce mainenance friction, would it be possible/safe to just run the default config of the validation?  what do we loose by doing that?

LuisPH3 (commented):
The commit could not be validated, check your keys.

.golangci.yml
facuMH: there are over 30 undocumented subpackages and over 100 variables that need renaming (most if not all of them going from `userId` to `userID` for example). In other projects we have chosen to not adhere to these rules because fixing them would not improve much on readability.

.github/workflows/go.yml
facuMH: https://github.com/0xsoniclabs/norma/pull/201/commits/63b2fccc70cda64002818fcc66865d15c6ad7cdc

.golangci.yml
facuMH: This file does not require further maintenance and it helps maintain the code style throughout the different repos. 

after an offline discussion with @LorenzSchueler we decided to keep the config file.

LuisPH3 (approved):

PR #203, opened 2026-06-11
Add bls-add scenario to validate allegro

simonlechner:
This PR adds a new app and scenario that calls the bls12-addition pre-compile with a valid input. Versions that do not support this pre-compiled contract which has been introduced in Allegro produce a different block hash.


LuisPH3 (approved):
thanks!

PR #204, opened 2026-06-11
Use slog or t.log instead of fmt.print.

facuMH:
This PR replaces uses of `fmt.Println` or `fmt.Printf` and replaces with either `slog.` or `t.Logf` as better fitting to each location.
The aim of these changes is to get a more homogenous log when running scenarios with Norma.

### DIFF
--- a/driver/node/opera.go
+++ b/driver/node/opera.go
@@ -22,6 +22,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"log/slog"
 	"os"
 	"regexp"
 	"slices"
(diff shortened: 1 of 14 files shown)


driver/node/opera.go
```diff
@@ -213,11 +214,11 @@ func StartOperaDockerNode(ctx context.Context, client *docker.Client, dn *docker
 func printLog(node *OperaNode) error {
 	reader, err := node.StreamLog()
 	if err != nil {
-		return fmt.Errorf("cannot read node logs: %e", err)
+		return fmt.Errorf("cannot read node logs: %w", err)
 	}
 	scanner := bufio.NewScanner(reader)
 	for scanner.Scan() {
-		fmt.Printf("[Opera Node %s] %s\n", node.GetLabel(), scanner.Text())
+		slog.Info("[Opera Node]", "label", node.GetLabel(), "message", scanner.Text())
```
LuisPH3: please keep this one as printf, because otherwise it will nest logs withing logs and the format will be quite unreadable.

LuisPH3 (approved):

facuMH:
superseded by https://github.com/0xsoniclabs/norma/pull/206

PR #205, opened 2026-06-11
Add option to automatically open report

LuisPH3:
Add `--open-report` flag to norma run command. The scenario report will automatically be open in the system configured browser.

### DIFF
--- a/driver/norma/run.go
+++ b/driver/norma/run.go
@@ -22,6 +22,7 @@ import (
 	"io/fs"
 	"log/slog"
 	"os"
+	"os/exec"
 	"os/signal"
 	"path/filepath"
 	"syscall"
@@ -55,6 +56,7 @@ var runCommand = cli.Command{
 		&skipChecks,
 		&skipReportRendering,
 		&outputDirectory,
+		&openReport,
 	},
 }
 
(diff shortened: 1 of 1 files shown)


driver/norma/run.go
```diff
@@ -283,3 +295,14 @@ func runScenario(ctx context.Context, path, outputDir, label string, keepPrometh
 
 	return nil
 }
+
+func openBrowser(s string) error {
+
+	path, err := exec.LookPath("xdg-open")
+	if err != nil {
+		return fmt.Errorf("xdg-open not found: %w", err)
```
facuMH, line 304:
```suggestion
		return fmt.Errorf("xdg-open failed to open: %w", err)
```

facuMH (dismissed):

driver/norma/run.go
LuisPH3: no, this command checks that xdg-open exists.

facuMH: but if xdg exists and failed to open the url because there is no browser installed, it would read:
````
xdg-open not found: 
/usr/bin/xdg-open: 882: x-www-browser: not found
/usr/bin/xdg-open: 882: firefox: not found
/usr/bin/xdg-open: 882: iceweasel: not found
/usr/bin/xdg-open: 882: seamonkey: not found
/usr/bin/xdg-open: 882: mozilla: not found
```

facuMH (approved):

PR #209, opened 2026-06-12
Add sending tx retry

thaarok:
Seems txs registering node/advancing epoch/etc are sometime lost from the txpool:
```
ERROR[01-01|00:02:48.672] event execution failed time=115.6 name="[node-latest-validator-0] Creating node" event_time=60.0 error="failed to register validator node; failed to register validator node; failed to create validator, receipt error: failed to get transaction receipt: timeout" duration=53.239s
```
This PR adds automatic retry while waiting for the tx receipt.

### DIFF
--- a/driver/network/backend.go
+++ b/driver/network/backend.go
@@ -21,7 +21,6 @@ import (
 	"strings"
 
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
-	"github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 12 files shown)


LuisPH3:
Replaced with other mechanism. the goal is to avoid this from happening in the first place, we need guarantees that norma can change rules or advance epochs for the scenarios to be meaningful testing tools.

PR #213, opened 2026-06-15
Single proposer blackout

simonlechner:
This PR adds a minimal blackout scenario similar to release tests case 19. 
A new blocks halted check is added to ensure no more blocks are produced once the second validator is offline.


simonlechner:
Scenario structure has been restructured, this scenario will be ported to the new syntax.

PR #216, opened 2026-06-16
Reduce norma memory use

LuisPH3:
This PR reduces container memory use from ~700 MB to ~350 MB. 

Old images not having access to these command line arguments are removed from testing.


facuMH (approved):
Since version are changed, should you remove also `2.1.5` ?

LuisPH3:
> Since version are changed, should you remove also `2.1.5` ?

I added it to be able to write scenarios during the last hardfork. just for completeness. this allows having scenarios which reproduce the hardfork @ 2.1.6 as a regression. 
I added it to this list so that the creation and the argumets are tested somewhere.

PR #224, opened 2026-06-19
Add testify dependency

LuisPH3:
To use require and assert in tests.

### DIFF
--- a/load/app/app_test.go
+++ b/load/app/app_test.go
@@ -168,6 +168,9 @@ func TestGenerators_Subsidies(t *testing.T) {
 		NetworkRules: rules,
 	})
 	require.NoError(t, err, "failed to create new local network")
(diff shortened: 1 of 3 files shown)


load/app/app_test.go
```diff
@@ -176,40 +169,23 @@ func TestGenerators_Subsidies(t *testing.T) {
...
 		NetworkRules: rules,
 	})
-	if err != nil {
-		t.Fatalf("failed to create new local network: %v", err)
-	}
-	t.Cleanup(func() {
-		if err := net.Shutdown(); err != nil {
-			t.Fatalf("failed to shutdown network: %v", err)
-		}
-	})
```
LorenzSchueler, line 186: What about `net.Shutdown()`?

LorenzSchueler (approved):

PR #225, opened 2026-06-19
Fix norma transaction tipping

LuisPH3:
Previous patch would modify values on error. could panic.


simonlechner (dismissed):
As discussed offline, the new gas values are fine for norma scenarios.

simonlechner (approved):

PR #226, opened 2026-06-19
Remove jenkins parallel limit

LuisPH3:
Since images are now smaller, tests do not parallelize apps docker instances. 
the rest of the project can be run with all available resources.


LuisPH3:
parallel is to be reused, limits are needed

PR #227, opened 2026-06-19
Add rules applied check

LuisPH3:
This PR adds a check which queries the rules from all nodes and verifies that the expected rules are applied.

### DIFF
--- a/driver/executor/sequential.go
+++ b/driver/executor/sequential.go
@@ -516,12 +516,12 @@ func execUpdateRules(step *parser.Step, net driver.Network) error {
 
 // checkFunctionToCheckerName maps check step functions to their checker names.
 var checkFunctionToCheckerName = map[parser.StepFunction]string{
-	parser.FuncCheckBlockGasRate:   "block_gas_rate",
-	parser.FuncCheckBlockHashes:    "blocks_hashes",
-	parser.FuncCheckBlockHeights:   "block_height",
-	parser.FuncCheckBlocksHalted:   "blocks_halted",
-	parser.FuncCheckBlocksProduced: "blocks_rolling",
-	parser.FuncCheckNetworkRules:   "network_rules",
+	parser.FuncCheckBlockGasRate:   "blockGasRate",
+	parser.FuncCheckBlockHashes:    "blocksHashes",
+	parser.FuncCheckBlockHeights:   "blockHeight",
+	parser.FuncCheckBlocksHalted:   "blocksHalted",
+	parser.FuncCheckBlocksProduced: "blocksRolling",
+	parser.FuncCheckNetworkRules:   "networkRules",
 }
 
 // execCheck runs a named checker with configuration from the check spec.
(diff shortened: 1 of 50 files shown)


driver/parser/sequential_test.go
```diff
@@ -571,11 +573,11 @@ Scenario:
 `,
 		},
 		{
-			name: "ceiling on checkBlocksProduced",
+			name: "ceiling on checks step",
 			input: `
 Name: Test
 Scenario:
-  - checkBlocksProduced:
+  - checks:
```
LuisPH3, line 580: wrong

driver/executor/sequential.go
```diff
@@ -511,8 +514,18 @@ func execUpdateRules(step *parser.Step, net driver.Network) error {
 	return nil
 }
 
-// execCheck runs a named checker with optional configuration from the step.
-func execCheck(ctx context.Context, checkerName string, step *parser.Step, checks checking.Checks) error {
+// checkFunctionToCheckerName maps check step functions to their checker names.
+var checkFunctionToCheckerName = map[parser.StepFunction]string{
```
LuisPH3, line 518: to PascalCase

scenarios/test/network_brio_hardfork.yml
```diff
@@ -29,14 +29,15 @@ network_rules:
...
+          Allegro: true  
           Brio: true  # hardfork enabled
 
 nodes:
     - name: node-v212
       start: 30
       failing: true
       client:
-        imagename: sonic:v2.1.2 # this node should stop working after the hardfork
+        imagename: Sonic:v2.1.2 # this node should stop working after the hardfork
```
facuMH: remove this change please.

scenarios/test/bundles_activation_coherence.yml
```diff
@@ -35,6 +35,7 @@ network_rules:
     - time: 50
       rules:
         Upgrades:
+          Allegro: true
```
facuMH, line 38: what should happen when brio is enabled but allegro is not ?

facuMH (changes requested):

scenarios/test/bundles_activation_coherence.yml
LuisPH3: there is an error reported. the rules are checked and if you write an scenario without the two it would not pass the unit test checking syntax 

this is a little annoying because you have to specify both every time, you cannot enable allegro in one update and brio later. but for now it is alright. we can make it more detailed later

driver/parser/sequential_check.go
```diff
@@ -174,3 +171,26 @@ func (s *Step) checkUpdateRules() error {
...
+		// Validate that rules patches are valid when provided.
+		if check.Function == FuncCheckNetworkRules && check.Rules != (genesis.NetworkRulesPatch{}) {
+			if err := genesis.ValidateNetworkRulesPatch(check.Rules); err != nil {
+				errs = append(errs, fmt.Errorf("sub-check %d (%s): invalid rules: %w", i+1, check.Function, err))
+			}
+		}
+	}
+
+	return errors.Join(errs...)
+}
```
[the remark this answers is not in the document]
simonlechner: Scenarios are written by developers, they should know what they are checking.

facuMH (approved):

PR #229, opened 2026-06-22
Remove container before test

facuMH:
When a unit test or scenario needs to start a new docker container with a sonic instance, `StartOperaDockerNode` checks whether the name is available and if not it fails. This was assumed as a safe method for debugging scenarios, but it assumed that docker runtime is clean, and that whichever test/scenario started a container, had the proper cleanup configured to remove it. 
This is not always the case when tests get interrupted or scenarios fails.

Hence this PR replaces the previously introduced check for existing containers with the name, for a force remove of the container, so that the new instance can start fresh.

### DIFF
--- a/driver/docker/docker.go
+++ b/driver/docker/docker.go
@@ -440,6 +440,25 @@ func (c *Client) ContainerExists(name string) (bool, error) {
 	return len(containers) > 0, nil
 }
 
+// ForceRemoveContainer force-removes a container with the given name if it
+// exists. This is useful for cleaning up leaked containers from previous test
+// runs before starting a new container with the same name.
+func (c *Client) ForceRemoveContainer(name string) error {
+	containers, err := c.cli.ContainerList(context.Background(), container.ListOptions{
+		All:     true,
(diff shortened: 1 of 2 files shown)


LuisPH3 (dismissed):

driver/docker/docker.go
```diff
@@ -440,6 +440,25 @@ func (c *Client) ContainerExists(name string) (bool, error) {
...
+// runs before starting a new container with the same name.
+func (c *Client) ForceRemoveContainer(name string) error {
+	containers, err := c.cli.ContainerList(context.Background(), container.ListOptions{
+		All:     true,
+		Filters: filters.NewArgs(filters.Arg("name", fmt.Sprintf("^/%s$", name))),
+	})
+	if err != nil {
+		return err
+	}
+	for _, cont := range containers {
```
LuisPH3, line 455: nit: add a warning: this could be stopping a concurrent test or and create weird interactions.

LuisPH3 (approved):

facuMH:
replaced by https://github.com/0xsoniclabs/norma/pull/230

PR #231, opened 2026-06-22
Remove old sonic version from scenarios

facuMH:
Since the new parameter flag was added in https://github.com/0xsoniclabs/norma/pull/216, running scenarios with older versions fails. 
This PR removes older version sonic from testing scenarios.


LuisPH3:
Overseen by https://github.com/0xsoniclabs/norma/pull/236

Scenarios are to be rewritten.

PR #244, opened 2026-06-29
Remove prometheus image.

facuMH:
This removes the `prom/prometheus:v2.44.0` Docker container that was spun up during scenario runs. The container was a leftover that served no functional purpose for Norma's metrics pipeline.

**Why this is safe?**

Norma collects Prometheus-format metrics directly from each Sonic node's built-in endpoint — it does not query the standalone Prometheus container.

The actual metrics collection happens in [PrometheusLogDispatcher](https://github.com/0xsoniclabs/norma/blob/e8b80f5/driver/monitoring/prom_log_provider.go#L104-L138), which HTTP GETs each node's `/debug/metrics/prometheus` endpoint every 1 second.

This dispatcher feeds all Prometheus-sourced metrics (txpool_*, system_cpu_procload, system_memory_used, db_size, statedb_disksize, emitter_*) into the monitoring pipeline via `prom_log_source.go`. None of these paths reference the standalone Prometheus container.
Other metrics are collected via monitoring and parsing of the logs, these mechanisms are all unaffected by the removed prometheus image.


scripts/run_sonic.sh
```diff
@@ -61,7 +61,7 @@ fi
 # round-trip time, the one-way latency has to be half of it.
 # To check, run `docker exec <src-container-id> ping <dst-container-id>` on host.
 echo "NETWORK_LATENCY=${NETWORK_LATENCY}"
-if [[ -n "${NETWORK_LATENCY}" ]]; then
+if [[ -n "${NETWORK_LATENCY:-}" && "${NETWORK_LATENCY}" != "0" && "${NETWORK_LATENCY}" != "0s" ]]; then
```
LuisPH3: is this alright?

facuMH: this change does not belong here. removed.

LuisPH3 (approved):

PR #246, opened 2026-06-30
Add default end of test checks to sequential scenarios

LuisPH3:
This PR adds the end of scenario checks to sequential scenarios syntax
introduced to scenarios in Norma for release testing 2.2

These checks advance epochs at the end and check that the nodes are all up to date. 
the option to disable these checks is given for tests which have special end conditions.


simonlechner (dismissed):

simonlechner (approved):
Nice refactor to use require.

PR #247, opened 2026-06-30
Node lifecycle and parallelization

simonlechner:
This PR removes the special treatment of validators specified in the beginning of the scenario. 
Node creation and deletion is parallelized for faster test times.

### DIFF
--- a/driver/executor/sequential.go
+++ b/driver/executor/sequential.go
@@ -27,6 +27,7 @@ import (
 	"github.com/0xsoniclabs/norma/driver/checking"
 	"github.com/0xsoniclabs/norma/driver/parser"
 	"github.com/0xsoniclabs/norma/genesis"
+	"golang.org/x/sync/errgroup"
 )
 
 // RunSequential executes a sequential scenario on the given network.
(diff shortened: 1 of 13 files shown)


driver/executor/sequential.go
```diff
@@ -282,47 +290,87 @@ func execStartNode(
...
+				instanceNames[instance], node, validatorIds[instance],
+			}
+			return nil
 		})
-		if err != nil {
-			return fmt.Errorf("failed to create node %s: %w", instanceName, err)
-		}
+	}
+	if err := g.Wait(); err != nil {
+		return err
```
facuMH: maybe expand the error "failed waiting for nodes to be created" or something similar.

driver/executor/sequential.go
```diff
@@ -282,47 +290,87 @@ func execStartNode(
 	// We'll wait for new nodes to reach this height before proceeding.
 	targetBlock, err := getNetworkBlockHeight(ctx, net)
```
LuisPH3, line 291: i think this may no longer needed, nodes are not created in this file anymore, therefore no waiting for block needed

LuisPH3 (dismissed):

driver/monitoring/network/validator_stakes.go
```diff
@@ -80,7 +80,9 @@ func newValidatorStakeSourceWithCollector(
 			case now := <-ticker.C:
 				stakes, err := res.collector()
 				if err != nil {
-					slog.Error("failed to fetch validator stakes", "metric", ValidatorStake.Name, "error", err)
+					if !errors.Is(err, driver.ErrEmptyNetwork) {
```
facuMH, line 83: in what cases does this function get called with an empty network ?

LuisPH3: it does get called each second, if the network has no nodes active it rendered an ugly warning.
This is a new case which could not be tested before.... all nodes are offline and the network comes back to life.

facuMH (approved):

LuisPH3 (approved):

PR #252, opened 2026-07-02
Remove port forwarding

facuMH:
This PR removes the port forwarding mechanism and replaces it by using the docker network assigned ips for the containers.


driver/node/opera.go
```diff
@@ -169,6 +171,17 @@ func StartOperaDockerNode(ctx context.Context, client *docker.Client, dn *docker
 		return nil, fmt.Errorf("failed to start docker node: container %q already running", config.Label)
 	}
 
+	// A custom bridge network is required so the host can reach the
+	// container by its IP. Create one if the caller did not provide it.
```
LuisPH3: what is the use case case of having a dedicated network and bridge network.
I am a little confused by these comments (here and in other code locations). 

The way I see it, one norma execution shall always spawn a custom network (which is bridged) and this network is an architectural piece of norma, I cannot imagine why any code path in norma could not be using it.

can you explain this problem? can you figure out why there are two or more networks, if they are needed and if this can be simplified?

facuMH: There are 5 different tests, which do not spawn their own networks, yet try to set up a node and wait for it to be alive. Instead of adding a network to those tests, and any test in the future which does not need a network, I thought it would be best to have a backup network. Hence if no network is provided the, one is created.
Tests such as https://github.com/0xsoniclabs/norma/blob/main/driver/monitoring/node/cpu_profile_test.go#L29 or https://github.com/0xsoniclabs/norma/blob/main/driver/node/opera_test.go#L63

LuisPH3: I think that instead increasing complexity, by allowing this layer to decide whenever a norma-net has or has not a docker-network, every test shall be migrated and the simples solution in the infrastructure shall be provided:
- every norma network spawns a bridged dedicated network. this is automatic and neither APIs or developers need to be concerned about it.

We should not merge this PR yet, the intention was to reduce complexity and this is not clear to me.

driver/node/opera_test.go
```diff
@@ -296,3 +310,19 @@ func TestClient_Stop_Graceful(t *testing.T) {
 		t.Errorf("container did not stop gracefully")
 	}
 }
+
+func createTestNetwork(
```
LuisPH3: this is a pretty elegant solution to the problem

driver/docker/docker.go
```diff
@@ -78,11 +95,11 @@ type ContainerConfig struct {
 	ImageName       string
 	ShutdownTimeout *time.Duration
 	Environment     map[string]string
-	Entrypoint      []string // Entrypoint to run when starting the container. Optional.
-	Network         *Network // Docker network to join, nil to join bridge network
-	DataDirBinding  *string  // mount client datadir to this path on host
-	GenesisFileBind *string  // mount genesis file on host to /genesis.json:ro in container
-	KeystoreBinding *string  // mount keystore dir on host to /datadir/keystore:ro in container
+	Entrypoint      []string      // Entrypoint to run when starting the container. Optional.
+	Network         DockerNetwork // Docker network to join, nil to join bridge network
```
LuisPH3: nil does not have special meaning anymore, right?

LuisPH3 (dismissed):

LuisPH3 (approved):

PR #254, opened 2026-07-03
Add event throttler check

simonlechner:
This PR adds a check to verify the event throttler works as expected.

### DIFF
--- a/driver/checking/event_throttled.go
+++ b/driver/checking/event_throttled.go
@@ -20,30 +20,31 @@ import (
 	"context"
 	"fmt"
 	"log/slog"
-	"math"
-	"math/big"
-	"sort"
+	"time"
 
 	"github.com/0xsoniclabs/norma/driver"
 	"github.com/0xsoniclabs/norma/driver/monitoring"
 	"github.com/0xsoniclabs/norma/driver/rpc"
-	"github.com/0xsoniclabs/sonic/gossip/contract/sfc100"
-	"github.com/0xsoniclabs/sonic/opera/contracts/sfc"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
 )
 
-// defaultThrottleCeiling is the default maximum ratio (in percent) between
-// the least-emitting non-dominant validator and the max-emitting dominant
-// validator. Values above the ceiling indicate the throttler is not
-// effective enough.
-const defaultThrottleCeiling = 50
+// minGapRatio is the minimum ratio required between the slowest
+// unthrottled validator and the fastest expected-throttled validator.
+// Below this ratio the split is considered too narrow to conclude that
+// throttling is occurring.
+const minGapRatio = 2.0
 
-// defaultDominantStakeThreshold matches sonic's emitter throttler default
-// (see emitter/config.DefaultThrottlerConfig). The dominant set is the
-// smallest set of highest-staked validators whose cumulative stake meets
-// or exceeds threshold * total stake.
-const defaultDominantStakeThreshold = 0.75
+// defaultSampleWindow is the interval between the two DAG snapshots used
+// to compute per-validator emission rates. Longer windows produce more
+// stable rate estimates but are more likely to straddle an epoch
+// boundary and force a retry.
+const defaultSampleWindow = 5 * time.Second
+
+// maxSampleAttempts caps the number of sampling attempts before giving
+// up. An attempt is discarded whenever the epoch rolls over between the
+// two snapshots that make up the window.
+const maxSampleAttempts = 5
 
 func init() {
 	RegisterNetworkCheck("eventThrottled",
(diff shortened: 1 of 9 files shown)


driver/checking/event_throttled.go
```diff
@@ -0,0 +1,337 @@
...
+
+// dagEvent represents a single DAG event with its creator and parents.
+type dagEvent struct {
+	creator uint64
+	parents []common.Hash
+}
+
+// collectEpochEventCounts fetches all events in the current epoch
+// via the DAG API and returns per-creator event counts.
+func collectEpochEventCounts(ctx context.Context, client rpc.Client) (map[uint64]int, error) {
```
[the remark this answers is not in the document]
simonlechner: Not an issue for norma scenarios.

simonlechner:
In order to run the added example scenario, #253 is required.

scenarios/sequential/throttler_check.yml
```diff
@@ -0,0 +1,32 @@
...
+    users: 30
+    rate:
+      constant: 15
+
+  - waitFor: 45s
+
+  - checks:
+      - blocksProduced:
+          tolerance: 10
+      - eventThrottled:
```
LuisPH3, line 33: I am a little confused by the check declaration:
there are 3 validators, who is throttling? All of them cannot be emitting 10% of events. 
What is the expectation which is checked?

driver/checking/event_throttled.go
```diff
@@ -0,0 +1,337 @@
...
+			minNonDom = count
+			minCreator = creator
+		}
+		if count > maxNonDom {
+			maxNonDom = count
+			maxCreator = creator
+		}
+	}
+
+	if nonDomCount < 2 {
```
LuisPH3: this is an odd expectation. with the appropiate stake, a network of 2 nodes can be dominated by one single node, this would make 1 nonDom

driver/checking/event_throttled.go
```diff
@@ -0,0 +1,337 @@
...
+			"total", total,
+			"percent", fmt.Sprintf("%.1f%%", pct),
+		)
+	}
+}
+
+// verifyThrottling checks that the least-emitting non-dominant
+// validator produced at most ceiling% of the most-emitting
+// non-dominant validator's events.
+func verifyThrottling(counts map[uint64]int, labels map[int]string, ceiling int) error {
```
LuisPH3: this function thinks the non dominant validator is the one which emitts the least. but it is the one with nonsignificant stake. 
I think that this function should deduce dominant set from stake and not emissions. 

the dominant set can be deduced using the functions in sonic.

LuisPH3 (changes requested):
I think the check for throttling is quite fishy. expects >= 2 suppressed nodes which is odd. 
it does not contemplate stake on the stakes distribution. 

It may be that this is testing something meanigfull, but i cannot tell. the specification of the check is a little odd as well.

driver/checking/event_throttled.go
simonlechner: No longer required.

driver/checking/event_throttled.go
simonlechner: Stakes are now queried from sonic.

scenarios/sequential/throttler_check.yml
simonlechner: The check has been refactored.

driver/parser/sequential.go
```diff
@@ -550,15 +629,17 @@ var checkFunctionParams = map[StepFunction][]string{
...
 	FuncCheckNetworkRules:   {"rules", "failing"},
 }
 
 // checkParamDescriptions provides a human-readable description for each sub-check parameter.
 var checkParamDescriptions = map[string]string{
-	"ceiling":   "Maximum allowed value (float64) for a gas rate check.",
-	"failing":   "When true, the check is expected to fail; a passing result is treated as an error.",
-	"rules":     "Expected network rules patch (NetworkRulesPatch field structure).",
-	"tolerance": "Allowed deviation (int, in blocks) between nodes for a height/production check.",
+	"ceiling":        "Maximum allowed threshold (float64); interpretation depends on the check (e.g. gas rate cap, or emission percentage).",
```
LuisPH3: i think ceiling does not read very well. it would be better to use a dedicated field (like tolerance)
it is not clear to me neither what percentage means when throttling, "at most ceiling % from the other"? 
maybe we just delete the parameter, is there a use case for this parameter? do we have an scenario where we require different values for ceiling?

simonlechner: Ceiling has been removed.

driver/checking/event_throttled.go
```diff
@@ -0,0 +1,453 @@
...
+
+	rates, err := c.collectRates(ctx, client, c.sampleWindow)
+	if err != nil {
+		return err
+	}
+
+	logEmissionRates(rates, labels)
+	logThrottledSet(expected, labels)
+
+	return verifyThrottled(expected, rates)
```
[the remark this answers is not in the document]
simonlechner: If a validator does not emit any events it is offline and not just throttling, it should not be counted as a throttled validator.

LuisPH3 (approved):

PR #257, opened 2026-07-06
Suppress error when node re-syncs

LuisPH3:
An error would print when a node is stopped and started again with the same name, blocks will be produced during synccing, because the metrics would already had accounted them before.
this PR introduces a synccing flag so that this error does not produce an error.

### DIFF
--- a/driver/monitoring/node/transactions_throughput.go
+++ b/driver/monitoring/node/transactions_throughput.go
@@ -19,7 +19,6 @@ package nodemon
 import (
 	"fmt"
 	"log/slog"
-	"sync"
 	"time"
 
 	"github.com/0xsoniclabs/norma/driver/monitoring"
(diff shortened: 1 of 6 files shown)


driver/monitoring/node/transactions_throughput.go
```diff
@@ -82,7 +85,43 @@ func (s *TransactionsThroughputSource) OnBlock(node monitoring.Node, block monit
...
+	}
+	return syncing
+}
+
+func (s *TransactionsThroughputSource) shouldSuppressAppendConflict(node monitoring.Node, err error) bool {
+	if !monitoring.IsOutOfOrderAppendError(err) {
+		return false
 	}
+	return s.isNodeSyncing(node)
 }
```
facuMH: should this be abstracted into a separate type which both block and tx_throughput use ?

facuMH (dismissed):

driver/monitoring/node/transactions_throughput.go
LuisPH3: factored

simonlechner (approved):

PR #259, opened 2026-07-07
Re-add single proposer blackout scenario

simonlechner:
This PR adds the fixed single proposer blackout scenario to the new release tests.
The `blocksRolling` check is extended with a `start` option, allowing to specify what timeframe should be considered by the check.

### DIFF
--- a/driver/checking/blocks_rolling.go
+++ b/driver/checking/blocks_rolling.go
@@ -3,6 +3,7 @@ package checking
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/0xsoniclabs/norma/driver"
 	"github.com/0xsoniclabs/norma/driver/monitoring"
@@ -20,7 +21,10 @@ func init() {
 type blocksRollingChecker struct {
 	monitor          MonitoringData
 	toleranceSamples int
-	start            monitoring.Time
+	// duration, when > 0, switches the checker into live-observation mode:
+	// Check blocks for that long and then verifies that at least one node
+	// advanced its block height during the observation window.
+	duration time.Duration
 }
 
 // Configure returns a deep copy of the original checker.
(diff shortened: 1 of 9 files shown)


driver/checking/blocks_rolling.go
```diff
@@ -58,9 +61,17 @@ func (c *blocksRollingChecker) Check(ctx context.Context) error {
 	// The test ensures that at least one node is generating blocks, confirming that the network is operational
 	// to some extent. It does not verify the functionality of every node, as that is handled by other checks.
 	//
-	// To account for flexibility in block processing, the verification is performed within a sliding window
-	// defined by 'toleranceSamples'. Only the block height at the beginning and end of this window are assessed,
-	// allowing for scenarios where blocks may not be produced every second.
+	// Two evaluation modes are supported:
```
LuisPH3: I would like that we change the semantics of this check.
Having the test operating retrospectively forces the person writing the scenario to care about the timing of the previous operations, this can yield time-dependent issues and is a little like the old timestamps thinking.
Having tests (this one and other time-window tests) considering only forward windows make tests a little easier to write:
the semantics should be, from this time for the duration specified, check that blocks are rolling. 

This will be more reliable, imagine that you have wait for epoch and then this check, for how long shall you measure? do you need to add a wait?
Implementation could be as easy as changing the field documentation and adding an implicit wait at the beginning of the check. 
start 0 disappears, I do not see much use in this one nevertheless

simonlechner: This version already uses a duration argument (starting from the time of the check) rather than the old start time which was before the check.
Removing the `duration == 0` option will lead to a lot of old cases failing. I will remove it with the old cases.

LuisPH3 (dismissed):

LuisPH3 (approved):

PR #261, opened 2026-07-13
Remove old scenario syntax

simonlechner:
This PR cleans up the norma codebase by removing the old scenarios as well as all code parts related to it.
The code should be reviewed by commit.


LuisPH3 (approved):
The replace here remove the old time based syntax for scenarios with the new sequential events syntax.  This is now the canonical norma script syntax.

PR #264, opened 2026-07-20
Exit on check failure and timeouts

simonlechner:
This PR improves the behavior of norma in case of a failure or timeout. 
Several functions did not handle a canceled context correctly.
Checks now exit the scenario in case of a failure.

### DIFF
--- a/load/app/context.go
+++ b/load/app/context.go
@@ -52,15 +52,22 @@ type RpcClientFactory interface {
 }
 
 // NewContext initializes an application context bound to a random RPC client,
-// treasury account, and the scenario network rules patch.
-func NewContext(factory RpcClientFactory, treasury *Account, networkRules genesis.NetworkRulesPatch) (AppContext, error) {
+// treasury account, and the scenario network rules patch. The provided context
+// is used for the network operations performed through this app context, so
+// cancelling it (for example on a check failure or timeout) aborts pending
+// receipt waits and RPC calls.
+func NewContext(ctx context.Context, factory RpcClientFactory, treasury *Account, networkRules genesis.NetworkRulesPatch) (AppContext, error) {
+	if ctx == nil {
+		ctx = context.Background()
+	}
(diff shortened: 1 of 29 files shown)


load/app/context.go
```diff
@@ -52,15 +52,22 @@ type RpcClientFactory interface {
...
 
 // NewContext initializes an application context bound to a random RPC client,
-// treasury account, and the scenario network rules patch.
-func NewContext(factory RpcClientFactory, treasury *Account, networkRules genesis.NetworkRulesPatch) (AppContext, error) {
+// treasury account, and the scenario network rules patch. The provided context
+// is used for the network operations performed through this app context, so
+// cancelling it (for example on a check failure or timeout) aborts pending
+// receipt waits and RPC calls.
+func NewContext(ctx context.Context, factory RpcClientFactory, treasury *Account, networkRules genesis.NetworkRulesPatch) (AppContext, error) {
+	if ctx == nil {
```
LuisPH3, line 60: If no context would be an error, it would be more difficult to have the wrong context.

LuisPH3 (changes requested):

load/app/context.go
```diff
@@ -57,10 +57,6 @@ type RpcClientFactory interface {
 // cancelling it (for example on a check failure or timeout) aborts pending
 // receipt waits and RPC calls.
 func NewContext(ctx context.Context, factory RpcClientFactory, treasury *Account, networkRules genesis.NetworkRulesPatch) (AppContext, error) {
-	if ctx == nil {
```
LuisPH3: this is not what i meant.
a nil context is still valid, go will treat it as background.
we do not want that happening because it hides timeouts and cancelations, if ctx is nil here, we shall return an error.

LuisPH3 (changes requested):

LuisPH3 (approved):

PR #265, opened 2026-07-20
Poll network rules check until all nodes converge

simonlechner:
The `networkRules` check took a single instantaneous snapshot of each node's applied rules. But a rules update applies per node at the epoch seal, and `waitForEpoch` only waits for one (random) node to advance, so immediately afterward a slower node can still report the pre-update rules.

### DIFF
--- a/driver/checking/network_rules.go
+++ b/driver/checking/network_rules.go
@@ -8,6 +8,7 @@ import (
 	"reflect"
 	"sort"
 	"strings"
+	"time"
 
 	"github.com/0xsoniclabs/norma/driver"
 	"github.com/0xsoniclabs/norma/driver/monitoring"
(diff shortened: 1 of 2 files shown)


driver/checking/network_rules.go
```diff
@@ -65,6 +78,36 @@ func (c *networkRulesChecker) Check(ctx context.Context) error {
...
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-ticker.C:
+		}
+	}
+}
+
+func (c *networkRulesChecker) checkOnce(ctx context.Context) error {
 	nodes := c.net.GetActiveNodes()
 	slog.Info("checking applied network rules for nodes", "count", len(nodes))
```
LuisPH3: nit: move logging outside retry, one print per check not per try

driver/checking/network_rules.go
```diff
@@ -65,6 +78,36 @@ func (c *networkRulesChecker) Check(ctx context.Context) error {
...
+	for {
+		err := c.checkOnce(ctx)
+		if err == nil {
+			return nil
+		}
+		if time.Now().After(deadline) {
+			return err
+		}
+		if !logged {
+			slog.Info("network rules not yet applied on all nodes; waiting for convergence", "timeout", c.timeout)
```
LuisPH3: too verbose, make it slog.Debug if you need it

LuisPH3 (changes requested):

driver/checking/network_rules.go
simonlechner: You are right, the check already prints a log message.

driver/checking/network_rules.go
simonlechner: Removed

LuisPH3 (approved):

PR #272, opened 2026-08-03
Ensure validators are actual validators

simonlechner:
This PR ensures that when a node is started as a validator, it is part of the active validator set by advancing the epoch.


LuisPH3 (dismissed):

simonlechner (commented):
Extra changes look good.

LuisPH3 (approved):

PR #282, opened 2026-08-12
Add priority lanes, a check that they work, and scenarios

LuisPH3:
Sonic's transaction priorities feature hoists the transactions of registered senders ahead of the rest at block formation, up to a gas quota per registered entity. Norma had no way to place load in such a lane.

### The lane

A lane is not a load of its own — the point is comparing the same traffic with and without a priority. The new `priority` application type therefore takes the load it carries as a parameter:

```yaml
- runApp: fast
  type: priority
  load: uniswap
```

It registers the accounts its users sign with as one entity and generates that load unchanged.

Loads whose senders are not known before the run refuse to be prioritized. Registering what such a load happens to hold at setup time would prioritize part of its traffic and silently leave the rest behind, which is worse than refusing.

### The check

`prioritizedTransactionsFirst` reads block order rather than latency. Latency is not deterministic here: it depends on the offered rate, the pool depth and the machine, and the advantage disappears once a lane saturates its quota — while the ordering stays correct throughout.

The check notes the head of the chain when it starts, waits for the blocks produced from there, and fails if:

- too few of them carry transactions of both classes — no evidence, rather than a pass on a network where nothing was prioritized;
- the run of prioritized transactions a block opens with covers less than a median share of the transactions the quota can admit (0.5 by default, lower it with `minRunCoverage` for a congested lane);
- the gas of that run exceeds the quota of one entity — asserted only where the block demoted nothing, since a demoted transaction sits among the ordinary ones and cannot be told from one.

The yardstick is what the quota admits, not every prioritized transaction in the block. A lane offering more gas than its quota necessarily leaves some behind, so the stricter yardstick would fail exactly when the quota does its job.

### The scenarios

| Scenario | Intent |
| --- | --- |
| `examples/priority_lanes.yml` | Show the effect at all: a lane, the identical unprioritized load, and congestion. |
| `release_testing/.../lane_ahead_of_ordinary_load.yml` | A lane below all of its limits. |
| `release_testing/.../lane_under_transaction_count_load.yml` | A lane under load measured in transactions. |
| `release_testing/.../lane_under_heavy_gas_load.yml` | A lane under load measured in gas, pinned at its quota. |
| `load_generators/priority.yml` | Smoke test for the new application type. |

Each release test asserts the ordering with the new check.

Stacked on #281.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/SCENARIO_SPECIFICATION.md
+++ b/SCENARIO_SPECIFICATION.md
@@ -237,10 +237,38 @@ Starts a load-generating application. The value is the application’s
 ```
 
 **Supported application types** (case-insensitive): `counter`, `erc20`,
-`store`, `uniswap`, `smartaccount`, `subsidies`, `transient`,
+`store`, `uniswap`, `smartaccount`, `subsidies`, `priority`, `transient`,
 `selfdestructoldcontract`, `selfdestructnewcontract`, `ecdsa`,
 `largecontract`, `allofbundle`, `oneofbundle`, `subsidizedbundle`,
 `failingbundle`, `duplicatedbundle`, `bls12add`, `mix`.
+
+A priority lane is a property of a load rather than a load of its own, so
+`priority` does not generate traffic itself: the optional `load` parameter names
+the type that does, and the accounts signing it are registered in the on-chain
+priority registry.
+
+```yaml
+- runApp: fast
+  type: priority
+  load: uniswap            # optional; any type above, defaults to counter
+  rate:
+    constant: 20
+```
+
+Running that next to a plain `uniswap` application isolates the effect of the
+feature, since the two loads differ in nothing else. `load` applies to `priority`
+only; on any other type it is an error, as is nesting one lane in another.
+
+`priority` requires the `TransactionPriorities` upgrade to be active and fails
+to start otherwise; see
+[scenarios/examples/priority_lanes.yml](scenarios/examples/priority_lanes.yml)
+for a scenario making the effect of the lanes visible.
+
+The loads whose users create the accounts they sign with while they run - the
+bundle types and `mix` - cannot be prioritized: their accounts are not known when
+the registry is written, and registering the ones that happen to exist would
+prioritize a part of their traffic and quietly leave the rest behind. Asking for
+them reports that rather than doing it.
 
 **Rate shapes** — exactly one of the following must be set on `rate`:
 
@@ -359,6 +387,7 @@ Upgrades:
   SingleProposerBlockFormation: <bool>
   GasSubsidies: <bool>
   TransactionBundles: <bool>
+  TransactionPriorities: <bool>
 ```
 
 Type notes:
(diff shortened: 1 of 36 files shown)


LorenzSchueler (commented):
An AI generated PR deserves an AI generated review 😉 

## Correctness

### 1. The quota assertion is vacuous in the one scenario built to exercise it

`driver/checking/priority_ordering.go:276` skips the `overQuota` accounting whenever
`order.run < order.prioritized`. But demotion *is* what the quota binding produces: the client
hoists what fits and appends the rest in base order.

In `scenarios/release_testing/features/priority_lanes/lane_under_heavy_gas_load.yml` the store
transactions cost `52000 + 25000*260 = 6,552,000` gas each against the 100M per-entity quota
(~15 per block), while the lane offers 20/s. So essentially every mixed block shows demotion and
the quota is never asserted — yet the scenario's headline expectation is "the lane to be pinned at
its per-entity gas quota".

`TestPriorityOrdering_TheQuotaIsNotAssertedWhereTransactionsWereDemoted` documents this as
intended, so it is a design gap rather than a defect. Either state it in the scenario, or assert a
bound that survives demotion:

```go
// The prefix cannot be delimited exactly, but it cannot exceed the quota by more
// than the one transaction that ended it.
for entity, gas := range order.runGas {
    if gas > quota+order.maxRunGas[entity] {
        ...
    }
}
```

### 2. `run` is not the hoisted prefix when the prefix is empty — a false-failure path

`driver/checking/priority_ordering.go:350` counts the leading prioritized transactions, which
equals the client's hoisted prefix only when that prefix is non-empty.

If every prioritized sender is nonce-blocked, the block is plain base order. Should that order
happen to start with prioritized transactions *and* place all of them before all ordinary ones,
then `run == prioritized`, the demotion guard at line 276 stays silent, and `runGas` — unbounded,
since nothing was actually hoisted — can trip

> prioritized transactions were hoisted beyond the gas one entity may have per block

for a block that hoisted nothing. Unlikely under the shipped loads (needs a full nonce-block plus a
lucky base order), but it is a false failure rather than a missed one.

### 3. The RPC client is never closed

`driver/checking/priority_ordering.go:377-390` caches the dialled client on the source for the
process lifetime and closes it only on the registry-bind error path. Every other checker dials per
use and defers a close:

- `driver/checking/block_height.go:176`
- `driver/checking/event_throttled.go:143`
- `driver/checking/validators_active.go:149`

`Configure` copies the checker struct but shares the `source` pointer, so the connection survives
reconfiguration too.

---

## Consistency

### 4. `duration` missing from the check's parameter table

`SCENARIO_SPECIFICATION.md` §5 lists `minMixedBlocks`, `minRunCoverage`, `failing`. But
`checkFunctionParams[FuncCheckPrioritized]`, `PrintHelp` and the §5.1 prose all also accept
`duration`.

### 5. Scenario comment contradicts its config

`scenarios/examples/priority_lanes.yml:33` — "# Two validators, so that …" sits above
`instances: 3`. The sibling release scenario says "Three, so that …".

### 6. The quota rationale contradicts the heavy-gas scenario

`load/app/priority_app.go` calls the limits "deliberately generous: a load generator should be
limited by the network, not by a lane budget it happens to share with itself", while
`lane_under_heavy_gas_load.yml` is built to pin the lane at exactly that budget. One of the two
comments is stale.

---

## Minimization and duplication

### 7. The load validation is duplicated verbatim

`load/app/factory.go:33-47` and `driver/parser/check.go:178-187` carry the same three conditions
with near-identical messages, and are tested twice
(`TestNewApplication_RejectsALoadTheTypeCannotCarry` and `TestScenario_ApplicationLoadIsValidated`).
One `func ValidateLoad(appType, load string) error` in `load/app`, called from both, removes the
drift risk.

### 8. `NewPriorityApplication` duplicates the factory's default

It exists only for `load/app/app_test.go` and encodes "the default load is counter" a second time,
with a *different* justification in its doc comment than the factory's — "the load whose
transactions are the cheapest to distinguish from one another" vs "the cheapest traffic to compare a
lane against". Drop it and have the test call `NewApplication("priority", "", …)`.

### 9. Unrelated whitespace change

`driver/parser/check.go:43` — a blank line was removed before `// Validate each step.`

### 10. Redundant scenario parameter

All three release scenarios pass `minMixedBlocks: 10`, which is the default. They read as tuned when
they aren't.

### 11. `AllCheckFunctions()` aliases the package array

It returns `allCheckFunctions[:]`, so a caller can overwrite the registry. `slices.Clone(…)` is
safer for an exported accessor.

---

## Tests

### 12. `networkPrioritySource` has no test at all

It holds the riskiest code in the check — the anonymous `eth_getBlockByNumber` struct and its JSON
tags, the internal-transaction predicate, the `Level == 0` classification, the `IsUint64` guard on
the quota — and every test goes through `fakePrioritySource`. A wrong tag or a flipped predicate
ships silently.

At minimum, add a test decoding a canned `eth_getBlockByNumber` payload, and reference
`internaltx.IsInternal` in the comment at `driver/checking/priority_ordering.go:434` so the two stay
in sync.

### 13. Table-driven convention

`driver/checking/priority_ordering_test.go` is 15 functions, most sharing one shape: build a block,
maybe set a quota or config, assert pass or an error substring. That is a table.

`TestNewApplication_RejectsALoadTheTypeCannotCarry` also duplicates the three cases of the
already-table-driven `TestScenario_ApplicationLoadIsValidated`, and
`TestPriorityOrdering_TheQuotaIsCountedPerEntity` bundles a pass and a fail case in one function via
a re-assigned `source`.

### 14. Bare assertions

New tests attach messages, against the convention:

- `load/app/priority_app_test.go` — `"types are case-insensitive"`, `"the %s load is its own, so it
  takes no load parameter"`, `"the error should name the load that was asked for"`
- `driver/checking/priority_ordering_test.go` — `"four blocks are below the default requirement"`
- `driver/executor/run_test.go` — `require.Truef` / `require.Containsf`

PR #290, opened 2026-08-17
Skip nodes with a suspended client when managing peers

ihtefyw:
Peer operations retry an unreachable RPC endpoint for 180 seconds before giving up. A node whose client was deliberately taken down (SuspendNode, e.g. by a killSonic step) fed that retry loop anyway, so restarting one of several concurrently killed validators stalled for minutes per dead peer and pushed db-heal scenarios past the 10-minute deadline.

The network now remembers which nodes are suspended and skips them when adding or removing peers; they re-establish their connections themselves when resumed. Joining a network whose remaining nodes are all suspended is no longer an error, and a failing RemovePeer call no longer double-unlocks the nodes mutex.

### DIFF
--- a/driver/network/local/local.go
+++ b/driver/network/local/local.go
@@ -55,7 +55,14 @@ type LocalNetwork struct {
 	// validator nodes created during startup.
 	nodes map[driver.NodeID]*node.OperaNode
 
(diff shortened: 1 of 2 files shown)


simonlechner (approved):
In case a node is offline it has to be removed from the dial random rpc set.

PR #291, opened 2026-08-19
Log each scenario step under its own name, coloured per operation

ihtefyw:
Currently, in a norma scenario log each line are colored the same. This makes it difficult to distinguish between the network progression and the operations written in the log. 
This PR makes it easier by coloring some operations.

A step is now logged under the operation's own name, with the phase carried as an attribute:

    INFO [11:59:59.992] healDb    step=7 phase=started   identifier=frail
    INFO [12:00:00.240] healDb    step=7 phase=completed duration=248.047105ms

The terminal handler pads the message, so this turns that column into an index of the steps a run performed, and each operation is given its own colour to make it scannable. The `function=` attribute is dropped, as the message now carries it.

The new log looks like this: 
image

### DIFF
--- a/driver/executor/run.go
+++ b/driver/executor/run.go
@@ -83,6 +83,13 @@ func RunAndCaptureEventExecution(
 	return executions, err
 }
 
+// A step is logged twice, under the operation's own name, once as it starts and
+// once when it is done. These distinguish the two.
+const (
+	phaseStarted   = "started"
+	phaseCompleted = "completed"
+)
+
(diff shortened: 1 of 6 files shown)


simonlechner (approved):
Nice addition. I think readability could further improve if the two phase types (started, executed) were also colored.

ihtefyw:
@simonlechner added
image

LuisPH3 (approved):

PR #292, opened 2026-08-19
Send system transactions through a single node

ihtefyw:
Two advanceEpoch steps in a row could fail the second one with a receipt timeout reporting the transaction as "not present", which is what the end-of-scenario checks appended to every scenario do.

Epoch advances and rule updates are both sent from the shared system account, so they form one nonce sequence, and neither sets a nonce explicitly: bind reads it from the pending state of whichever node the call dialed. DialRandomRpc chooses a fresh node every time, and nodes observe a block a moment apart, so the second call could read a nonce the first had already spent. Such a transaction can never be included; it is eventually dropped from the pool, and the wait for its receipt times out with the transaction no longer there.

Observed mid-run, moments after the first advance was mined, as the pending nonce of the system account across the four nodes of a scenario:

    nonce = 10  10  10  11
         stable-0  -1  -2  frail

Sending these transactions through one pinned node rules the reordering out. Each waits for its own receipt from that node, so by the time the next one asks for a nonce, the node demonstrably holds the block containing the previous transaction and reports the nonce following it. The pin is re-chosen only when the node it names stops answering, which is also the one moment the hazard can return, since the sequence has to continue somewhere.

Read-only uses of DialRandomRpc are left alone; only the two operations sending from the shared account are pinned.

### DIFF
--- a/driver/network/local/local.go
+++ b/driver/network/local/local.go
@@ -94,6 +94,14 @@ type LocalNetwork struct {
 	// appContextMu guards lazy initialization of appContext.
 	appContextMu sync.Mutex
 
+	// systemRpcLabel is the node that transactions from the shared system
+	// account are sent through. See dialSystemRpc for why they must not be
+	// spread over several nodes. Guarded by systemRpcMu.
+	systemRpcLabel string
+
+	// systemRpcMu guards systemRpcLabel.
+	systemRpcMu sync.Mutex
+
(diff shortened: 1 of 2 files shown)


LuisPH3 (approved):
the pr adds infra to use the same entry-point for norma system txs.

possible improvements:
- always choose a validator node (avoid gossiping of sys txs)
- avoid "failing" nodes

PR #293, opened 2026-08-19
Keep the treasury's nonce in one sequence

ihtefyw:
The treasury the load framework pays for setup with is the same on-chain account the driver sends epoch advances and rule updates from, and every one of those transactions takes its nonce from the pending state of the node it happens to be sent through. Nothing kept those readings in one order, so two of them could pick the same nonce; the loser can never be included, is dropped from the pool, and the wait for its receipt times out reporting the transaction as absent. This is the hazard fixed for the driver's own transactions in the previous commit, reached from the other side.

An app context now shares the node those driver transactions are pinned to, so contract deployments and account funding join the same sequence rather than reading it from a node that may not have seen the latest of them yet. Generated load is unaffected: it comes from accounts of its own and still reaches the network through its own clients. Pinning only concerns transactions that spend the shared account's nonce.

FundAccounts also asked the node for a new nonce after submitting each batch, which reports the nonce that batch just used until the node takes it into its pool. The next batch could therefore reuse it and quietly replace its predecessor, leaving those accounts unfunded while both transactions appeared to have been sent. It now advances from the nonce it spent, which is the value it already holds, and one RPC per batch goes away with it.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

### DIFF
--- a/driver/network/local/local.go
+++ b/driver/network/local/local.go
@@ -95,7 +95,7 @@ type LocalNetwork struct {
 	appContextMu sync.Mutex
 
 	// systemRpcLabel is the node that transactions from the shared system
-	// account are sent through. See dialSystemRpc for why they must not be
+	// account are sent through. See DialSystemRpc for why they must not be
 	// spread over several nodes. Guarded by systemRpcMu.
 	systemRpcLabel string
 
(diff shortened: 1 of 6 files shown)


simonlechner (approved):

LuisPH3 (approved):
the pr assigns rpc endpoints to fund accounts from nodes which should be online

